PR #12, opened 2017-08-08
Remove composer.lock from git

svenluijten:
We don't need this in a package, the version constraints will make sure the right version(s) of dependencies are installed.


mattstauffer:
Thanks!

PR #13, opened 2017-08-08
Use phpunit.xml.dist

svenluijten:
It's common to provide a `phpunit.xml.dist` file to encourage the user to configure their own test environment. We'll provide sensible defaults though 🙂


DanielCoulbourne:
Thanks

PR #17, opened 2017-08-09
Add config file and route exclusion/inclusion logic

ctf0:
- <s>add auto discovery for 5.5</s> https://github.com/tightenco/ziggy/pull/20
- add config
- add support for excluding routes via whatever
- update rdme

<s>i would suggest to add an extra `/` at the start of the uri to avoid issues like segment duplication ex. `admin/admin/media`, not sure if this is a rare case but i believe it wont cuz any side effects if added.</s> **fixed**

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,19 +6,23 @@
 
 Ziggy creates a Blade directive which you can include in your views. This will export a JavaScript object of your application's named routes, keyed by their names (aliases), as well as a global `route()` helper function which you can use to access your routes in your JavaScript.
 
-## Installation 
+## Installation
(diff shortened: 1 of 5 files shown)


ctf0:
using the config will happen sooner or later, but i understand ur pov about keeping the package as simple as possible.

for the naming, honestly i should've wrote something more descriptive than `update` but i was too lazy.

u can keep this PR as a point of reference, and i will send another for the auto-disc

DanielCoulbourne:
Thanks! Closing this for now.

Issue #22, opened 2017-08-09
Add an artisan command to output the javascript to a file for inclusion


mattstauffer:
In case someone wanted to add this to their build process, we could have a command:

```bash
php artisan ziggy:generate
```

which saves the output of the script into a file wrapped in an importable module.

We could even include a sample watcher so your build tool watches the routes files for changes and then re-runs the command when necessary.

ctf0:
easier solution would have a config path of the js file, and the package watch that file and update it on the fly when app routes are changed, now the user only add the file to the build and forget about it.

mattstauffer:
Great thinking @ctf0. The reason I didn't want to do that is because this requires the user to use Mix or whatever else, which limits them a bit. Rather, I'd love a solution that works in a variety of settings AND ALSO makes it super easy to use with mix/elixir.

Additionally, people can have custom routes files, so whatever we do would have to be helplessly configurable.. or just a guideline, which is my suggestion.

ctf0:
i honestly wish if we dont use config files at all, but to keep things flexible we'll have to use the config like it or not, 

also you dont need to use any extra tool to use the js file, if you want you can simply call it with `{{ assets('js/ziggy.js') }}` just make sure the saved file path is under the public folder.

mattstauffer:
Not using config would be great.

The extra tool would be for watching the routes files and triggering the regeneration of the JS file.

ctf0:
extra tool, u mean `mix, elixir, etc.. ?` if yes then actually we can use something more clever and its within php, here is an exmaple https://github.com/ctf0/AssetCacheBust/blob/master/src/CacheBust.php

mattstauffer:
Right.. but you have to run, and configure, that command. 

Think through *when* this would be running.

Something, somewhere, has to notice every time routes.php has changed, and trigger re-generation of the javascript file. What is that something?

ctf0:
the only thing i can think about atm is an event that fires on each update to the routes list, then re-compile the js file.

or use a scheduler but not ideal.

mattstauffer:
@ctf0 you can't do that. There's nothing in the framework itself that *watches* the routes files. *something* has to trigger a comparison. If you wanted you could write that watching so that every web request checked it, but that would miss a bunch of key workflows.

Scheduler is a bad idea because most people don't run cron locally.

[...]

PR #25, opened 2017-08-11
auto compile js file when laravel routes are changed

ctf0:
https://github.com/tightenco/ziggy/issues/22

### DIFF
--- a/src/ZiggyServiceProvider.php
+++ b/src/ZiggyServiceProvider.php
@@ -2,6 +2,7 @@
 
 namespace Tightenco\Ziggy;
 
+use Illuminate\Support\Facades\File;
 use Illuminate\Support\Facades\Blade;
 use Illuminate\Support\ServiceProvider;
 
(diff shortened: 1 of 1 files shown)


mattstauffer:
Three problems here: 

First, the modified time on a folder doesn't change when the contents of its child files are changed.

Second, this slows down the entire app by adding this check every time the app is loaded.

Third, this won't update the js file in time if it's needed by a build tool; I could change all of my routes file and never open the app and this would never run, so the build tool would never get the updated javascript.

ctf0:
> First, the modified time on a folder doesn't change when the contents of its child files are changed.

actually on my mac the routes folder time gets changed every time i change any file under it, and i even have nested folders for each route group.

> Second, this slows down the entire app by adding this check every time the app is loaded.

it doesnt cost that much, but if u have a better idea, am all ears.

> I could change all of my routes file and never open the app and this would never run.

we can then have a cmnd that does that.

DanielCoulbourne:
@ctf0 I don't think we are going to merge this for a few reasons:

1) We don't want to have all these File checking operations appended to every request to the app. We are trying to minimize impact to load-time. In my mind one of the only reasons someone would opt to precompile this JS on their own is to avoid the minor performance hit of our blade directive. In that case, I think adding a bunch of file operations to the beginning of their app is kind of sketchy.

2) I think this is something that should only be fired in the local environment during development. I don't love the idea of tying into mix, but we already experimented with checking Route files for dirtiness and we opted to avoid that.

3) I think it's weird that we have a `src/js/route.js` file being compiled into PHP to be re-exported to another JS file. It just makes me feel weird personally.

ctf0:
actually this wont work for 2 reasons
1- watching the route folder is not effective, because what about the packages that register its own routes?
2- calling `app(BladeRouteGenerator::class)->generate()` outside of the blade dir, give empty array, not sure why through 🙄

PR #29, opened 2017-08-11
Re-work how base url is generated

### LINKED ISSUE #26: Return localhost instead of the contigured domain in virtual host
Hi I tried this package, but during the implementation of the helper function route() it returns http://localhost/..

For Example
```
  getData(route('ajax.get.event.list'), id, function(data) {
    data = data[0];
    data = editEventData(data);
    ...
```
the expected output should be 
`http://liz.dev/users/org-adviser/get/event`

the return instead is
`http://localhost/users/org-adviser/get/event`

mattstauffer:
Fixes #26 

Uses Laravel `route()` helper to take advantage of its default understanding of URL even if it's not set in `.env`

### DIFF
--- a/src/BladeRouteGenerator.php
+++ b/src/BladeRouteGenerator.php
@@ -16,15 +16,15 @@ class BladeRouteGenerator
 
     public function generate()
     {
-        $json = (string) $this->nameKeyedRoutes();
-        $appUrl = rtrim(config('app.url'), '/') . '/';
+        $json = $this->nameKeyedRoutes()->toJson();
+        $appUrl = url('/') . '/';
         $routeFunction = file_get_contents(__DIR__ . '/js/route.js');
 
         return <<<EOT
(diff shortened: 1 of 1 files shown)


src/BladeRouteGenerator.php
```diff
@@ -16,15 +16,15 @@ public function __construct(Router $router)
...
+        $json = $this->nameKeyedRoutes()->toJson();
+        $appUrl = url('/') . '/';
         $routeFunction = file_get_contents(__DIR__ . '/js/route.js');
 
         return <<<EOT
 <script type="text/javascript">
     var namedRoutes = JSON.parse('$json'),
         baseUrl = '$appUrl';
-    $routeFunction
+        $routeFunction
```
mattstauffer: Oops. @DanielCoulbourne didn't intend to introduce that. Sorry!

PR #30, opened 2017-08-11
Add support for unkeyed params array.

JesseLeite:
This PR adds support for:

- Unkeyed params array.
- Single non-array/object param.

This makes the JS route() helper a bit more consistent with Laravel's route() helper.  Example usage:

screen shot 2017-08-11 at 7 22 13 pm

I'm not sure where to get started regarding your test suite though.  Never used mocha or orchestra test bench.  If anyone is willing to help write/run tests, or willing to give me a few pointers, I'm more than down to giver a shot?

**UPDATE:**
Added test coverage for you 👍


DanielCoulbourne:
I love this. @mattstauffer mentioned the very same thing the other day. Thanks for the PR!
I'll dig in and try to get some tests written tonight or tomorrow. I'm not going to merge until we have tests running, but yeah I'm into it.

JesseLeite:
@DanielCoulbourne Thanks.  I actually have another improvement to allow this:

```js
route('posts.show', 123)
```

Because Laravel allows for a single param to be passed without array as well.

Think that's worth adding to this PR?  Or would you rather that be separate?

JesseLeite:
I added it anyway, it can always be reverted 😬

I'm not sure what you think of the use of `typeof` and/or `Array.isArray()`?  I'm not an expert with raw javascript type checking, and usually rely on lodash, etc. for that stuff.

Also, which podcast?

JesseLeite:
Read into mocha and added test coverage for my additions.  I think you might need to give codeship CI a kick though?

JesseLeite:
@DanielCoulbourne Thanks! I actually added one more little improvement to allow a single non-array/object param (as Laravel's `route()` allows for this as well).  Example...

```js
route('posts.show', 123)
```

I'm not sure what you think of the use of typeof and/or Array.isArray()? I used typeof in the last commit because it apparently returns object for both arrays and objects in JS. I'm not an expert with raw javascript type checking, and usually rely on lodash, etc. for that stuff.

Lastly, I read into mocha and added test coverage for my additions.  I think you might need to give codeship CI a kick though? 😬

JesseLeite:
Thanks! 👌

DanielCoulbourne:
Thank YOU sir 😄 

Nice, clean, tested PRs are always welcome around here!

I'll probably tag a release later tonight or tomorrow once I merge in a few more things, update the docs, etc.

PR #32, opened 2017-08-13
Don't pick methods property

ankurk91:
I could not find any usage of `methods` property, so removing it.
This will save some bytes on generated html page.


DanielCoulbourne:
Hey man. Thanks for the help, but I actually am planning to use those eventually.
I don't think the extra bytes are too bad TBH 😂

Issue #31, opened 2017-08-12
Optional parameters not supported


ankurk91:
Hi,

* ziggy : v0.2.2
* Laravel: v5.4.30

`route()` method throws error if [optional](https://laravel.com/docs/5.4/routing#parameters-optional-parameters) parameters not passed
```php
//routes/web.php
Route::get("/check/{id?}")->name("check")
```
```js
// `id` should be optional, next line will throw error
route("check")
// works when id is passed
route("check",{id:123})

// works when id is passed (master branch)
route("check",{"id?":123})
```

DanielCoulbourne:
Good call. I'll try to get this fixed soon.

DanielCoulbourne:
Closed by #33

PR #33, opened 2017-08-13
Fix optional route parameters

ankurk91:
### Highlights -
* Make optional route parameter optional

The solution is little ugly, but works well.
Feel free to [refactor](https://help.github.com/articles/checking-out-pull-requests-locally/).

### Here are some samples -
```php
// routes/web.php
Route::get('/test/{id}/{next}/{from?}')->name('test');
```
```js
//  will throw error about missing id parameter only
route('test')

//  works well, skipping `from` parameter
route('test', { id : 123, next : 'checkout'} );
// http://domain.dev/test/123/checkout

//  works well with all parameters
route('test', { id : 123, from : 'product', next : 'checkout' });
// http://domain.dev/test/123/checkout/product
```
Ref: #31

P.S. Add tests
Thanks.

### DIFF
--- a/src/js/route.js
+++ b/src/js/route.js
@@ -1,5 +1,5 @@
 var route = function(name, params, absolute) {
-    if (params === undefined) params = {}; 
+    if (params === undefined) params = {};
     if (absolute === undefined) absolute = true;
 
     var domain = (namedRoutes[name].domain || baseUrl).replace(/\/+$/,'') + '/',
@@ -10,12 +10,16 @@ var route = function(name, params, absolute) {
     return url.replace(
         /\{([^}]+)\}/gi,
         function (tag) {
-            var key = Array.isArray(params) ? paramsArrayKey : tag.replace(/\{|\}/gi, '');
+            var key = Array.isArray(params) ? paramsArrayKey : tag.replace(/\{|\}/gi, '').replace(/\?$/, '');
             paramsArrayKey++;
-            if (params[key] === undefined) {
-                throw 'Ziggy Error: "' + key + '" key is required for route "' + name + '"';
+            if (typeof params[key] !== 'undefined') {
+              return params[key].id || params[key];
             }
-            return params[key].id || params[key];
+            if (tag.indexOf('?') === -1) {
+              throw 'Ziggy Error: "' + key + '" key is required for route "' + name + '"';
+            } else {
+              return '';
+            }
         }
     );
 }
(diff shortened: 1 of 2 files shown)


DanielCoulbourne:
@ankurk91 Thanks for the contribution! If you don't mind would you write a few tests to prove that this works? Check out @JesseLeite 's PR for [an example.](https://github.com/tightenco/ziggy/pull/30)

Thanks!

DanielCoulbourne:
Thanks @ankurk91 !

If you don't mind fixing the merge conflicts and writing a few tests to prove that it works, I'm happy to merge it.

Check out [this PR](https://github.com/tightenco/ziggy/pull/30) by @JesseLeite for an example of how/where to write the tests if you're unfamiliar. [Mocha Documentation](http://mochajs.org/#assertions) will probably help as well.

Robert-Hansen:
Hi I just created the required tests but there seems to be a problem with the code in the route file because now some of the tests fail here is the list of those

1. Should return URL when running with single non-array/object param on a route with required params
2. Should return URL when running with single array param on a route with required params
3. Should return URL when running with multiple array params on a route with required params 
4. Should return URL when running with whole object params on a route with required params 
5. Should return URL when running with some whole object params on a route with required params 
6. Should return correct URL when running with params on a route with required domain params

ankurk91:
I am using a different approach here,
I split whole url into segment and loop (map) through them.
The existing code do a regex and loop.
What we can do is check if current segment is an optional field and don't throw any error.
```js
function (tag) {
 // checkif tag is optional
// if tag is passed in params then apply otherwise skip
}
```

DanielCoulbourne:
@Robert-Hansen Are your tests somewhere I can access them? I'll get all this sorted out tomorrow then merge.

ankurk91:
He made a PR here
https://github.com/ankurk91/ziggy/pull/1

DanielCoulbourne:
@ankurk91 I like this PR but I think I'm going to be breaking some things pretty soon related to this, I'll do my best to try to update this PR to be mergeable after I make the breaking changes tomorrow because you've done a bunch of work and I'd love for you to get the little `Contributor` badge.

ankurk91:
I will try to incorporate recent changes in my code this tomorrow.
Will also import test cases from -
https://github.com/Robert-Hansen/ziggy/tree/fix/optional-params

I am actively monitoring this repo because i believe this gonna be part of my all future projects. :)

ankurk91:
Updated PR with test s.
ping @DanielCoulbourne

Issue #46, opened 2017-08-17
Remove default parameters or setup babel?


zackify:
https://github.com/tightenco/ziggy/blob/8962ad96cda7daedd76bf9268861112b613d6063/src/js/route.js#L1 

[Default params](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) only work in modern browsers. For this to work in IE 11 or anything else before es6 you might want to use babel, or just remove the default params and add a check at the top of the function? :)

ankurk91:
May be generate a minified version along.

mattstauffer:
That's totally my fault. I'd rather not introduce babel just for this. I'll update. Thanks @zackify!

DanielCoulbourne:
I think we will eventually end up adding Babel as the Javascript becomes more in-depth, but I agree, lets hold off for now.

zackify:
@DanielCoulbourne if you want I could send a pr that adds minimal stuff and a precommit hook that builds it. It's whatev you want bro

ankurk91:
Params still have a default value set to `{}`.

mattstauffer:
uhguhghghghghghgh
thanks totally did not notice lol

PR #50, opened 2017-08-18
Fix more default params

mattstauffer:
because I'm an idiot and missed this one

#46


mattstauffer:
(wait, didn't push right)

Issue #16, opened 2017-08-08
Add ability to exclude routes


svenluijten:
Let's say I have some super secret endpoints (`/admin/whatever`), I'd like to be able to exclude them from the routes that are exposed via the generated JSON.

Maybe this can be achieved by adding a middleware, like how [`aaronlord/laroute`](https://github.com/aaronlord/laroute) does it. 

As an extension of this, it'd be neat if I could use route groups to only expose certain routes to a "named" group, so I could use `@routes('admin')` in my blade file(s) to only expose routes in the `admin` group.

DanielCoulbourne:
I've been doing some thinking along these lines as well.

mikemand:
I dislike the way `laroute` does this since I have to go through and blacklist any routes I don't want to be visible, and if I forget to blacklist a route...

I would much rather have to whitelist the specific routes (or a route group). I also like your idea of using the directive to only expose routes, but I wonder if it would be possible to make it so they have to fit a specific pattern (`admin.*` or `api.v1.*`)?

ctf0:
i think i can make a PR for this as i was just thinking of the same thing

DanielCoulbourne:
@mikemand I was initially planning to export by default and require a blacklist. Can you think of any situations where accidentally exposing a route would cause a serious issue? I don't see it as a security issue so far, but happy to be proven wrong.

I think I'm okay with using a whitelist, but I'd prefer to use groups. Let me do a little thinking on this before you start working on a PR @ctf0, I may want to do something with macroing Router or something. If you have any specific plans for implementation let me know :)

ctf0:
@DanielCoulbourne actually the PR is already made 😆 .

regarding the route group, i dont think laravel exposes this info through the `router->getRoutes()` however here is a way around if u r interested https://github.com/ctf0/pretty-routes/blob/master/views/routes.blade.php#L41-L51

DanielCoulbourne:
@ctf0 I made a few comments on the PR. In general I'd like to keep to one major change per PR. I would have loved to merge parts of it but not the whole thing. More info in my review of #17 

Thanks for being willing to write the code! Let's work together to get it merged 😄

mikemand:
@DanielCoulbourne I cannot really think of anything, security-wise. Any serious API should have ways to prevent DoS attacks and the like, but this doesn't just expose API routes; it exposes all of them.

[...]

PR #51, opened 2017-08-18
Optional whitelist groups

JesseLeite:
After some discussion in #16, here's my PR for optional whitelist groups.  It's tested and documented, but here's a quick rundown of my implementation:

If you add `groups` to your ziggy config file:

```php
return [
    'groups' => [
        'admin' => [
            'admin.*',
            'posts.*',
        ]
        'author' => [
            'posts.*',
        ]
    ],
];
```

You essentially have multiple whitelists.  Using the `@routes` blade directive without a parameter behaves as it always did.  However, you can pass a group key like `@routes('author')` to only expose the whitelisted routes in that specific group.  This can be helpful if you have multiple user roles and want more control over which routes you expose in each user dashboard.

**Note:** The existing `whitelist` and `blacklist` features still behave as expected, though if you expose a specific `group`, it will take precedence over `whitelist` and `blacklist` 👍 

Thoughts? 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,14 +66,14 @@ return axios.get(route('posts.show', post))
 ```
 
 ## Filtering Routes
-
 Filtering routes is *completely* optional. If you want to pass all of your routes to JavaScript by default, you can carry on using Ziggy as described above.
 
-If you do want to filter routes, we have provided two optional configuration settings to allow you to do so. To take advantage of these, create a standard config file called `ziggy.php` in the `config/` directory of your Laravel app and set **either** the `whitelist` or `blacklist` setting to an array of route names.
+### Basic Whitelisting & Blacklisting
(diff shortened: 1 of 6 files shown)


DanielCoulbourne:
I like this. I'll also review this in a few.
Thanks!

JesseLeite:
Btw, because I made a change to the blade directive, you might need to run `artisan view:clear`.  I'm not sure if this would be a problem for users when upgrading, do you know?

ankurk91:
Don't you think it should be -
```php
<?php
return [   
    'whitelist' => [
        'groups' => [
            'admin' => [
                'admin.*',
                'posts.*',
            ],
            'author' => [
                'posts.*',
            ]
        ],
    ]
];
```
Notice the `whitelist` key in array.

JesseLeite:
Mainly just trying to keep usage as simple as possible, which is why I avoided more complex nesting.  I thought it might add confusion around a single global whitelist, versus more specific groups of routes, ie) `@routes('author')`.

JesseLeite:
💞

PR #52, opened 2017-08-18
Pull in phpunit

JesseLeite:
Pull in phpunit instead of relying on global install?


JesseLeite:
Hmm that's embarassing.  I must've had something wrong with my vendor folder... orchestra/testbench pulls it in, and a composer re-install fixed my issue.

PR #53, opened 2017-08-19
Break router out into a class, including optional fluent `with()` method

DanielCoulbourne:
This is basically a housekeeping update to begin the transition to using object methods in the Router. This allows us to create fluent chainable methods on the base.

This PR implements `route().with()` which accepts all the same input as the `params` argument of the existing `route()` function.

It's purely cosmetic an exists as a proof of concept. I'm not going to document it yet, just in case we decide that fluency for core functionality is a bad idea.

Currently works like so: 
```javascript
route('team.user.show').with({team: 'tighten', id: 1})
```

I will soon add `withQuery` and `withDomain` methods as well. I still intend to fit queries into the base function arguments somehow to achieve feature parity with Laravel as best as possible, but I think this provides a nice alternative for folks who have a hard time parsing what is going on visually.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,8 @@
   },
   "homepage": "https://github.com/tightenco/ziggy#readme",
   "devDependencies": {
-    "mocha": "^3.5.0"
+    "axios": "^0.16.2",
+    "mocha": "^3.5.0",
+    "moxios": "^0.4.0"
   }
 }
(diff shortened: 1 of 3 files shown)


DanielCoulbourne:
I want to write a few more tests to verify that this works in a real word scenario before merging (want to make sure that the string gets returned when used as an axios argument, for example.)

JesseLeite:
I love the idea of fluency, so long as the core `route()` function can be used like Laravel.  This is beautiful for alternative/additional functionality though.  I'm honestly surprised that Laravel's `route()` doesn't return object and `__toString()`!

PR #57, opened 2017-08-21
some fixes

Robert-Hansen:
removed redundant escape
added missing var and semicolon
added test for unknown route and if route() is used with no route name
added check for unknown route and if you use route() with no route name provided it will also throw an error

@DanielCoulbourne


src/js/route.js
```diff
@@ -15,10 +12,15 @@ Router.prototype.toString = function() {
 };
 
 Router.prototype.constructDomain = function() {
-    if (this.absolute)
+    if(this.name === undefined) {
```
ankurk91: @Robert-Hansen 
Any reason behind removing check for absolute ?
```js
if (this.absolute)
```
https://github.com/tightenco/ziggy/blob/master/src/js/route.js#L18

DanielCoulbourne:
Thanks!

src/js/route.js
JesseLeite: I'm curious as well.  Maybe we should write a test to cover the absolute param and PR this back in.

DanielCoulbourne: Yeah I actually saw this after I merged. That was the reason for my wanting to revisit the tests before I tag a new release 😄 #61

PR #65, opened 2017-08-23
[0.4.0] Add test for axios query params

ankurk91:
Seems like i found the issue.

It is not working with `axios` when we pass `params` as second parameter
```js
axios.get(route('posts.index'), {
  params: { page: 2 }
})
```
This is just one case, there may be hidden scenarios.

I was able to fix it by calling ``toString()`` but this is not the best way to fix this.
 ```js
axios.get(route('posts.index').toString(), {
  params: { page: 2}
})
```
I think we should stick to return string type.

### DIFF
--- a/tests/js/test.route.js
+++ b/tests/js/test.route.js
@@ -147,6 +147,24 @@ describe('route()', function() {
         moxios.uninstall()
     });
 
+    it('Should make an axios call with query params when a route() is passed', function(done) {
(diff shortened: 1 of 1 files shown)


DanielCoulbourne:
@ankurk91 Thanks for finding the bug and being willing to write a PR, but I found a fix that lets me keep the object syntax. Check out #66

PR #66, opened 2017-08-23
[0.4.1] Add String passthrough to handle edge-cases.

DanielCoulbourne:
@ankurk91 This should fix the behavior you were seeing before. We are now extending the `String()` prototype so any unrecognized methods will be passed along to String.

Basically Router is now a `String` object with a few extra methods.

I added a test to demonstrate axios working with the params object you noted before. Thanks for finding that. The reason it was failing was that Axios uses `url.indexOf` when appending params. So because `Router` didn't have an indexOf method, it failed. Now that `Router` extends `String` it just passes unknown method calls on to String, which handles them.


JesseLeite:
Oh interesting!

PR #68, opened 2017-09-03
[WIP] Blacklist macro

davidianbonner:
Using a fluent syntax over a config would be my preferred way of black/whitelisting. 

Works as expected with routes and simple groups. 

`prefix`, `domain`, `middleware`, and `namespace`look to be the stumbling block.


DanielCoulbourne:
Thanks man! I'll check in on this later this week for sure. I'm a bit buried right now.

davidianbonner:
Evening folks.

After some playing around with this a ton of implementations in a relatively large app (200+ routes) I'm quite happy with this implementation. 

Simply put, instead of confusing matters with method chaining etc, it becomes a lot easier to follow if blacklisting is just a case of wrapping the routes in a group-esque manner:

```php
Route::blacklist(function () {
    Route::get('/posts',  'PostsController@index')->name('posts');
    Route::get('/posts/{id}', 'PostsController@show'); // No name – ignored by default
});

Route::get('/tags', 'TagsController@index')->name('tags');

Route::blacklist(function () {
    Route::get('/tags/{id}', 'TagsController@show')->name('tags.show');
});
```

Any thoughts on this implementation?

davidianbonner:
A cleaner and final implementation now committed.

Allows for blacklisting and whitelisting on single routes and groups of routes with less changes to the RoutePayload class.

```php
Route::blacklist()->get(...);
Route::whitelist()->get(...);

Route::blacklist(function () {
   Route::get(...);
});
Route::whitelist(function () {
   Route::get(...);
});
```

If support was only for 5.5 and up, we can improve on the single route listing because `Illuminate\Routing\Route` is macroable which would allow the below. I don't think having to declare blacklist and whitelist before anything else is that terrible though.

```php
Route::get(...)->name()->blacklist();
```

DanielCoulbourne:
What happens if you declare a macro on a non-macroable class? I'd love to implement that macro and if it breaks gracefully in 5.4 that would be cool :)

davidianbonner:
An exception is thrown (`FatalThrowableError`).

We could perform a conditional with `(float) app()->version() > 5.4` and register the macro only when this is true. `Route::get(...)->blacklist()` would fail if it was used. Though as long as the documentation stated that the fluent implementation worked in 5.5 and above I don't see this as a problem.

This feature doesn't feel like it warrants doing anything like extending the Route and Router classes?

Issue #74, opened 2017-09-22
Routes within a domain group do not work correctly


xandor:
## Missing Protocol

When using routes in a domain group, the domain is added by the js route() function, but there is no protocol attached, so the url returned is incorrect.

in web.php

```php
Route::domain('example.org')->group(function() {
	Route::get('/help', function () { 
			return 'help'; 
		})
		->name('help');
});
```

in blade/js

```js
route('help'); 
// returns 'example.org/help' 
//     instead of '//example.org/help' 
//             or 'http://example.org/help'

```

## Missing Port

Another thing is that if Laravel is running under example.org:81 the php route() function adds the :81 to the url, if its the current domain.

If the current URL is 'http://example.org:81/home'

```php
dump(route('help'));
// returns 'http://example.org:81/help'
```

but in JS

```js
route('help'); 
// returns 'example.org/help' 
//     instead of '//example.org:81/help' 
//             or 'http://example.org:81/help'
```

DanielCoulbourne:
Thank you for the excellent bug report!

I will address these soon!

DanielCoulbourne:
Just saw your PR. I’ll review and hopefully merge it this weekend.

DanielCoulbourne:
#75 Should be ready to merge after some minor changes.

xandor:
sorry i am a little bit under load at the moment, maybe can have a look this weekend.

xandor:
Thanks for the merge

PR #75, opened 2017-09-22
#74 fix domain bound routes

xandor:
add support for ports

### DIFF
--- a/src/BladeRouteGenerator.php
+++ b/src/BladeRouteGenerator.php
@@ -49,9 +49,10 @@ EOT;
     {
         $url = url('/');
         $parsedUrl = parse_url($url);
+
         $this->baseUrl = $url . '/';
-        $this->baseProtocol = array_key_exists(PHP_URL_SCHEME, $parsedUrl) ? $parsedUrl[PHP_URL_SCHEME] : 'http';
-        $this->baseDomain = array_key_exists(PHP_URL_HOST, $parsedUrl) ? $parsedUrl[PHP_URL_HOST] : '';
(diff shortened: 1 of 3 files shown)


src/BladeRouteGenerator.php
```diff
@@ -21,16 +27,35 @@ public function getRoutePayload($group = false)
...
+    $routeFunction
 </script>
 EOT;
     }
+
+    private function prepareDomain()
+    {
+        $url = url('/');
+        $this->baseUrl = $url . '/';
+        $this->baseProtocol = str_before($url, ':');
```
DanielCoulbourne: I think this would break in the case where protocol was `//` and there was a port.
For example – `//example.com:8080`

In this case, `baseProtocol` would end up being `//example.com`

I think you could get around that by doing `str_before($url, '//')` and removing the `:` if it exists.

src/BladeRouteGenerator.php
```diff
@@ -21,16 +27,35 @@ public function getRoutePayload($group = false)
...
+        $this->baseDomain = str_replace($this->baseProtocol . '://', '', $url);
```
DanielCoulbourne: You'll also have to update this concatenation for the `//` case.

DanielCoulbourne:
@xandor I'll have time to work on this later this week, but if you're able to make these changes real quick I can merge it now

src/BladeRouteGenerator.php
xandor: can `url('/');` return a string like '//example.org:8080'? 

I think it takes the url from the request and so it should always contain a complete protocol and not the shortcut version.

xandor:
@DanielCoulbourne I'm now using `parse_url` to analyze the url. that should be more robust...

PR #86, opened 2017-10-28
Refactor to use an ES6 Class with Babel and Mix.

DanielCoulbourne:


ankurk91:
Hi,
Nice to see the progress, Here are some thoughts


### Can we switch to 
```js
throw new Error('some error');
```
### dist js file is not minified 
It was intentional, right ?
If we minify the source we can use comments there which will help new contributors to understand whats going on there.

### Object.assign is not supported by IE
Would you like to configure a babel plugin ?

### Laravel mix already comes with babel-preset-env 
babel-preset-es2015 was intentional ?
We can stick to babel-preset-env and specify the [browsers](https://github.com/ai/browserslist) we want to support in `.babelrc`.

### Larevel mix is too large to compile just a single js file.
We can do this via [babel-cli](https://babeljs.io/docs/usage/cli/).
```
babel script.js --out-file script-compiled.js
```
OR switch to custom webpack setup.

### mix-manifest.json can be gitignored

Thanks

DanielCoulbourne:
Thanks for all the thoughts! Yes this was definitely my preliminary "Get it working" version.

I definitely want to get better build tooling set up, just used mix because it's so easy.
@ankurk91 you seem to know a lot more about babel than I do. If you wouldn't mind would you be open to setting up the build tools (or talking me through setting them up) including the polyfill for `Object.assign()`

PR #87, opened 2017-10-29
ES6, Webpack and much more

### LINKED ISSUE #79: Object.assign() is not supported in IE
I am unable to use Ziggy in IE, due to the [lack of support for Object.assign()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Browser_compatibility). When I attempt to use `route()` I get this:

screen shot 2017-10-17 at 9 41 51 pm

This can be solved with a [polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Polyfill), when necessary.

`Object.assign()` is used twice:

* [https://github.com/tightenco/ziggy/blob/master/src/js/route.js#L23](https://github.com/tightenco/ziggy/blob/master/src/js/route.js#L23)
* [https://github.com/tightenco/ziggy/blob/master/src/js/route.js#L50](https://github.com/tightenco/ziggy/blob/master/src/js/route.js#L50)

ankurk91:
This PR is based on #86 

### Highlights
* ES6 everywhere
* Webpack 
    - Minimal setup 
    - Generate two dist files, minified and non-minified
    - [babel-preset-env](https://babeljs.io/env/) and [browserlist](http://browserl.ist/?q=%3E+2%25%2C+last+2+versions) take care of browser support
   - Fixes #79 IE issue
   -  Watch mode for continuous build
   - `npm run build:watch`
* Mocha-webpack
   - Test case can use ES6 syntax
   - Build (run webppack) then test against `dist` files
   - Watch mode to see test results in realtime
   - `npm run test:watch`
* Specify node js version in package.json file
     - This will prevent `npm install` on older version and force same version across contributors machines
* Throw Error object `throw new Error()` instead of string
* Update .gitignore
     - ignore phpStorm/Webstorm generated files
     - ignore any *.log file (usually created by node)
* API remains the same
   - Users can still use `route()` method same as before
   - We are taking advantage of webpack [library](https://webpack.js.org/guides/author-libraries) build to expose `route` method to window object
  - Dist files are ready to use in module environment, like
  - `import route from 'dist/js/route.js'`
  - Dist files can be dumped to html page and works without any module bundler too
 * Speed up CI by caching vendor files

@DanielCoulbourne 
Feel free to checkout this PR [locally](https://help.github.com/articles/checking-out-pull-requests-locally/) and make your changes.


DanielCoulbourne:
Awesome I'll read this before bed! THANKS so much.

reinink:
Can we get a release tagged by any chance?

I have a project that uses IE11 and I can't use this project without this fix. 😭

DanielCoulbourne:
@reinink sorry man I meant to tag this sooner but wanted to rewrite the Docs first. I just tagged it and will clean up the docs in v0.5.1

reinink:
@DanielCoulbourne Thanks!

PR #97, opened 2017-11-14
Remove JSON.parse()

ankurk91:
I don't think we need `JSON.parse` here.
It works well without parsing.

Another approach could be like -
```html
<script type="text/javascript">
    window.Ziggy = @php echo json_encode([
      'property1' => $variable,
      'property2' => $variable2,
    ]); @endphp
  </script>
```


DanielCoulbourne:
Awesome.

PR #99, opened 2017-11-15
Environment based loading of route.js file

ankurk91:
Lets load non-minified file only when `APP_ENV = local` otherwise load `route.min.js`.


DanielCoulbourne:
Thanks for all 3 PRs man! Things are looking good!

the94air:
Guys... We should get this documented.

jakebathman:
Good idea, lemme write it up.

PR #100, opened 2017-11-16
Allow publishing config

ankurk91:
This PR will allow publishing default configs. User no need to manually create config file.

These configs will not be used by package unless user publish it.
We can publish config with this command:
```
php artisan vendor:publish --provider="Tightenco\Ziggy\ZiggyServiceProvider" --tag="config"
```

**todo**
* Documentation on how to publish config in readme.md
* Document each of the config property in `config/ziggy.php`, what they do.
* At-least add a note about using one either `blacklist` or `whitelist` property.

### DIFF
--- a/config/ziggy.php
+++ b/config/ziggy.php
@@ -1,0 +1,8 @@
+<?php
+
+return [
+
+    //'whitelist' => [],
+    'blacklist' => [],
+    'groups' => []
+];
--- a/src/ZiggyServiceProvider.php
+++ b/src/ZiggyServiceProvider.php
@@ -28,5 +28,10 @@ class ZiggyServiceProvider extends ServiceProvider
                 CommandRouteGenerator::class,
             ]);
         }
+
+        $this->publishes([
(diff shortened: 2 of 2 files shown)


DanielCoulbourne:
I'm still not really into publishing configs. Sorry haha.

JoshuaDoshua:
Any chance opinions on this has changed? I see configs everywhere, but no actual config file.

If its still:
> I'm still not really into publishing configs. Sorry haha.

Would you consider an alternative of just creating a config file in this package that sets the defaults so it can be easily referenced?

Happy to open a new PR or update this one; tagging @bakerkretzmar since it's an old Issue and you seem to be pretty active here

bakerkretzmar:
Sure, and I actually have a branch locally where I've played with this. I don't love it either but it is nice to have everything listed in one place. I'll try to work on it but I'd be open to a PR too 👍🏻

WalterWoshid:
> I'm still not really into publishing configs. Sorry haha.

Why, just why? Looking up the config in the docs or looking up the code in the vendor directory is just stupid.

PR #101, opened 2017-11-18
Register blade directive after resolving service

Omranic:
More efficient way (IMHO) to register the blade directive, to make sure it's only registered when `blade.compiler` service is resolved (we're browsing from web and we're rendering blade views already), otherwise we neither need to attach the directive nor resolve the `blade.compiler` at this stage. Using the `Blade::directive` facade will auto resolve the `blade.compiler` service even if we're interacting with the application from CLI through artisan, which in some project setups it may cause some `InvalidArgumentException` exceptions (Route [name] not defined.)

### DIFF
--- a/src/ZiggyServiceProvider.php
+++ b/src/ZiggyServiceProvider.php
@@ -2,10 +2,9 @@
 
 namespace Tightenco\Ziggy;
 
-use Closure;
 use Illuminate\Support\Facades\Route;
-use Illuminate\Support\Facades\Blade;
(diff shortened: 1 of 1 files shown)


ankurk91:
Thanks.!
I was about to submit the same PR.
I have seen this approach in other packages as well.
For example
https://github.com/spatie/laravel-permission/blob/master/src/PermissionServiceProvider.php#L58

@DanielCoulbourne
Can we get this merged and tag a new release?

PR #114, opened 2017-12-28
Update route.js

shpontex:
Default Values:
just add compatibility for Laravel defaults values :
https://laravel.com/docs/5.5/urls#default-values
just add  
var defaultParameters = {
//example            
locale: "{{ $locale }}"
}
before @routes if you have default parameters

### DIFF
--- a/src/js/route.js
+++ b/src/js/route.js
@@ -41,9 +41,17 @@ class Router extends String {
                 let keyName = tag.replace(/\{|\}/gi, '').replace(/\?$/, ''),
                     key = this.numericParamIndices ? paramsArrayKey : keyName;
                 
-                 if (defaultParameters[keyName]) {
-                    tags[key] = defaultParameters[keyName]
-                 }
+                    if (typeof defaultParameters !== 'undefined' && defaultParameters[keyName]){
+                        let this_key = key
+                        if(tags[this_key]){
+                            while(tags[key]){
+                                key++
+                            }
+                            tags[key] = tags[this_key]
+                        }
+                        tags[this_key] = defaultParameters[keyName]
+                        key = this_key
+                    }
                 
                 paramsArrayKey++;
                 if (typeof tags[key] !== 'undefined') {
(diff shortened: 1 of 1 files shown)


ankurk91:
Can you add tests?
All of our js routes are under `Ziggy` namespace, can you also use the same?

shpontex:
@ankurk91 
ok ill try to make the tests

DanielCoulbourne:
I'm going to close this for now. I'm not clear on exactly what this does and how.

If you'd like to re-open please open an issue describing what you think needs to be different and then write some tests showing that that works.

Thanks.

ankurk91:
I think this PR belongs to this somehow
https://github.com/laravel/framework/pull/22572

shpontex:
@DanielCoulbourne 
in laravel like mentionned here https://laravel.com/docs/5.5/urls#default-values
we can use default paramerters
for example 
Route::get('/{locale}/posts/{id}', function () {
    //
})->name('post.index');
for locale we have to put it every time in each route for all project , exp : route("post.index",["en",3])
with default parameters you can add this in middleware  URL::defaults(['locale' => $request->user()->locale]);
and now the route is =>  route("post.index",3)
so i wanted to add this to ziggy because for now i have to share my locale variable for all views to put it in the ziggy route
the code that i add allows you to add
var defaultParameters = {
//example
locale: "{{ $locale }}"
}
and now you can use route("post.index",3) in ziggy

shpontex:
@ankurk91 
yes its related to that PR
in laravel there is no way to get the default parameters
so insted of putting 
var defaultParameters = {
locale: "{{ $locale }}"
}
in ziggy 
we can get default parameters and put in directly inside de ziggy global varibale so it will be automatically loaded

shpontex:
@ankurk91 
taylor otwell has merged my getDefaultParameters function
so for the next laravel release we will be  able to not put 
var defaultParameters = {
//example
locale: "{{ $locale }}"
}
i will try to add the default parameters to ziggy global variable 
i will reopen this with a new code when the new laravel release show up

PR #116, opened 2017-12-29
Fixed wrong URL generation

glaubersp:
If deployed to a subfolder, baseURL is correctly set, but it's not used to generate the final URL.
I'm not sure why JS code needs to parse the URL, but in that case, the parsing function should be changed.


DanielCoulbourne:
@glaubersp Can you provide an example of your expectation of behavior vs. current behavior? 

Also, this code breaks several existing tests.

Before merging I'd like to see existing tests passing and new tests added to show that your use case works.

Thanks for contributing. Hit me up if you have questions or want to work through it.

glaubersp:
Hello Daniel,
Sorry about breaking the tests but I've never used mocha and I could debug your tests to check the needed changes.
I have a laravel app running under an apache subfolder (as in `/var/www/html/subfolder`) and it's accessible by the url `http://subdomain.domain.com/subfolder/`.
The variable `baseURL` is correctly set to `http://subdomain.domain.com/subfolder`when debbuging and thus it could be directly concatenated to the actual path.

I didn't understand all the tests you've done in your code of the `setDomain` function, but I saw that the test at line 22 should be true (`this.route.domain` should be false) in this case and it was not.

I assumed `baseURL` is being correctly set by the PHP code at `BladeRouteGenerator:62` (`$this->baseUrl = $url . '/';`), so I just returned this value.

As I could not follow the `setDomain` logic, this assumption may be wrong.

I have few experience with JS coding but I've tried changing your tests to use this Ziggy object:

```javascript
global.Ziggy = {
    namedRoutes: { "home": { "uri": "/", "methods": ["GET", "HEAD"], "domain": null }, "team.user.show": { "uri": "users/{id}", "methods": ["GET", "HEAD"], "domain": "{team}.myapp.dev" }, "posts.index": { "uri": "posts", "methods": ["GET", "HEAD"], "domain": null }, "posts.show": { "uri": "posts/{id}", "methods": ["GET", "HEAD"], "domain": null }, "posts.update": { "uri": "posts/{id}", "methods": ["PUT"], "domain": null }, "posts.store": { "uri": "posts", "methods": ["POST"], "domain": null }, "posts.destroy": { "uri": "posts/{id}", "methods": ["DELETE"], "domain": null }, "events.venues.show": { "uri": "events/{event}/venues/{venue}", "methods": ["GET", "HEAD"], "domain": null }, "events.venues.index": { "uri": "events/{event}/venues", "methods": ["GET", "HEAD"], "domain": null },"optional":{"uri":"optional/{id}/{slug?}","methods":["GET","HEAD"],"domain":null}},
    baseUrl: 'http://www.myapp.dev/subdomain/',
    baseProtocol: 'http',
    baseDomain: 'www.myapp.dev',
    basePort: false
};
```

and then I've manually changed all test cases to match the new url `http://www.myapp.dev/subdomain/`, but I wasn't able to fix the following test cases as they seem to depend on the logic implemented at `setDomain` that I couldn't grasp completely:

    1) Should return correct URL when run with params on a route with required domain params
    2) Should return correct URL without port when run with params on a route with required domain params
    3) Should return correct route name for current() and respond accurately when queried.
    4) Should respond "true" on route().current() regardless of trailing slashes or query variables.

I hope the explanation helps you.
If you need anything else to solve this problem, I'd be glad to help.
Cheers.

DanielCoulbourne:
Just realized all the changes in this PR are to the `dist` version of the JS file, not the `src` version. 
I'm going to close this PR for that reason, as well as the fact that this is untested and breaks existing tests.. I'll take a look at identifying and fixing the bug that caused this PR tomorrow.

PR #123, opened 2018-01-19
Add the ability to skip the inclusion of the route function

raphaelsaunier:
This PR contains a small change to the `@route` Blade directive that allows us to keep the dynamic generation of the routes while bundling the `route` helper function separately (instead of shipping an extra 5KB with every page).

I wrote this in the simplest, backward compatible way without any tests so don't hesitate to close this PR if you want to solve this in a completely different way. ;)

### DIFF
--- a/src/BladeRouteGenerator.php
+++ b/src/BladeRouteGenerator.php
@@ -31,7 +31,7 @@ class BladeRouteGenerator
 
         $json = $this->getRoutePayload($group)->toJson();
 
-        $routeFunction = file_get_contents($this->getRouteFilePath());
+        $routeFunction = $this->getRouteFunction();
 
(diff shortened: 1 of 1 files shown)


rjvandoesburg:
Nice change, would use.
Currently overwriting the class myself and removing the script. If only protected properties and methods were used I would only need to overwrite generate and not the whole class 👎

raphaelsaunier:
Yeah I was surprised that this wasn't done by default. Even when minified, that's an extra 5KB of JS that almost never changes and still needs to be downloaded and parsed when the page loads. And with the undocumented `php artisan ziggy:generate` you lose the dynamic generation feature.

the94air:
@jakebathman And also `php artisan ziggy:generate` as @raphaelsaunier said.

jakebathman:
Yep, I'm on it. Gonna write up that whole artisan command and some of the other missing docs.

the94air:
Thanks in advance for any help you are able to provide. 😄

PR #124, opened 2018-01-26
Add encodeURIComponent to path and query params.

### LINKED ISSUE #118: Encode URLs?
I ran into a situation today where I was using user-generated input to generate URLs with Ziggy. For example:

```js
route('user.search', { ...my search params... })
```

However, if a user types in any character that has a special URL meaning (ie. an ampersand), then this URL is busted.

For example, if I search for "Jonathan&Amy", I'll get:

```
/users?search=Jonathan&Amy
```

Since the ampersand is a query string separator, my `search` value will only be "Jonathan".

A simple solution here would be to simply encode the URL when generating it using `encodeURIComponent()`. More info here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent.

DanielCoulbourne:
Closes #118

### DIFF
--- a/src/js/route.js
+++ b/src/js/route.js
@@ -60,7 +60,7 @@ class Router extends String {
                 paramsArrayKey++;
                 if (typeof tags[key] !== 'undefined') {
                     delete this.queryParams[key];
-                    return tags[key].id || tags[key];
+                    return tags[key].id || encodeURIComponent(tags[key]);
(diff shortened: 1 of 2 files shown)


DanielCoulbourne:
Merging and removing Codeship from this repo because I hate it.

Issue #115, opened 2017-12-28
Routes Blade Directive not being recognized


rorymcdaniel:
I'm using Laravel 5.5.28 and have tightenco/ziggy 0.6.0 installed. Console output shows that tightenco/ziggy was in fact discovered.

I've placed @routes in my base layout blade template, but it's not being rendered. Looking at the source simply shows @routes there instead of the JavaScript output I would expect.

Am I missing something obvious? What's the best way to troubleshoot this?

Sirhaziq:
i am experiencing the same problem as well. someone should help please

ankurk91:
Does `@routes()` make any difference?

rorymcdaniel:
No, it's the same in either case. It simply puts @routes() in the final HTML output, as if Blade isn't recognizing the directive.

rorymcdaniel:
It looks like if I make a change in ZiggyServiceProvider it works. I replaced: 
`
        $this->app->afterResolving('blade.compiler', function (BladeCompiler $bladeCompiler) {
            $bladeCompiler->directive('routes', function ($group) {
                return "<?php echo app('" . BladeRouteGenerator::class . "')->generate({$group}); ?>";
            });
        });
` 
with:
`
        Blade::directive('routes', function($group = null){
            return "<?php echo app('" . BladeRouteGenerator::class . "')->generate({$group}); ?>";
        });
`
And then it works as expected. I'd be happy to do a pull request with that change if you'd like.

arturmamedov:
No for me with `laravel 5.4` this fix doesn't work.

ToujouAya:
same as me
it is not work in laravel 5.4.36 
tightenco/ziggy 0.6.0

ankurk91:
It is working fine on:
* PHP 7.1.10
* Laravel Framework 5.4.36
* Packages installed - `composer show`
```
bugsnag/bugsnag                       v3.10.0 Official Bugsnag notifier for PHP applications.
bugsnag/bugsnag-laravel               v2.11.1 Official Bugsnag notifier for Laravel applications.
bugsnag/bugsnag-psr-logger            v1.4.0  Official Bugsnag PHP PSR Logger.
composer/ca-bundle                    1.0.7   Lets you find a path to the system CA bundle, and includes a fallba...
dnoegel/php-xdg-base-dir              0.1     implementation of xdg base directory specification for php
doctrine/inflector                    v1.2.0  Common String Manipulations with regard to casing and singular/plur...
[...]

PR #127, opened 2018-02-08
Fix @routes Blade not working on some env (issue #115)

darron1217:
I made a pull request for issue #115

### DIFF
--- a/src/ZiggyServiceProvider.php
+++ b/src/ZiggyServiceProvider.php
@@ -18,10 +18,8 @@ class ZiggyServiceProvider extends ServiceProvider
             return Macro::whitelist($this, $group);
         });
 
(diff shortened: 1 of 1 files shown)


inxilpro:
If any other service provider that's loaded before ziggy resolves the `blade.compiler` during boot, then the `afterResolving` hook won't fire. This is a very common pattern, and is even in the [Laravel docs](https://laravel.com/docs/5.5/blade#extending-blade). This PR is a good fix for that.

darron1217:
@inxilpro Thanks for sharing information :)
I actually didn't understand why my fix is working.. haha

DanielCoulbourne:
I'm going to merge this so people can pull down dev if they want, but I'm going to wait until I figure this and a few other things out before tagging a new release.

DanielCoulbourne:
Thanks for finding a fix for this @darron1217

PR #138, opened 2018-03-11
Add example using query parameters

### LINKED ISSUE #134: Documentation about querystrings.
Hello, I think the documentation should be updated, because of #34 . In the documentation there is nowhere to be written about the querystrings. I think it would be nice to add the code what Jesse wrote, or something similar.
for example:

```
var user = {id: 1, name: "John Doe", email: "john@example.com"};
var store = {id: 2, name: "example store"};

route('user.store.period',
                    {   user: user,
                            store: store,
                            country: null,
                            from: "2017-08-01",
                            till: "2017-09-01"},
                    );

//example.org/user/1/store/2/period?country=&from=2017-08-01&till=2017-09-01
```
I don't know exactly if the code what I wrote here is correct, but this just for example.

jakebathman:
This fixes #134, adding a quick example showing how additional data is turned into query parameters in the URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,12 @@ route('events.venues.show', {event: 1, venue: 2}) // Returns '/events/1/venues/2
 route('events.venues.show', [1, 2]) // Returns '/events/1/venues/2'
 ```
 
+With query parameters:
+
+```js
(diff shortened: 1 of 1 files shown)


DanielCoulbourne:
Thanks man

PR #143, opened 2018-03-14
Handle situations where a single model object is passed into URL.

### LINKED ISSUE #96: Ziggy does not find the id key in object
Hi, 

using Ziggy v0.5.0 I have the following error.

capture d ecran 2017-11-12 a 20 28 43

It works fine if I directly pass an id.

DanielCoulbourne:
fixes #96


src/js/route.js
```diff
@@ -6,17 +6,28 @@ class Router extends String {
...
+        // If the tags object contains an ID and there isn't an ID param in the
+        // url template, they probably passed in a single model object and we should
+        // wrap this in an array. This could be slightly dangerous and I want to find
+        // a better solution for this rare case.
+        
+        if (params.hasOwnProperty('id') && this.template.indexOf('{id}') == -1) {
+            params = [params.id];
+        }
 
+        this.numericParamIndices = Array.isArray(params);
```
piotr-lottestyle: this is not working if there are route modal bindings and query paramaters (product/{product}/edit?someParam=123)

Issue #107, opened 2017-11-29
dingo/api Routing Support


bagaskarawg:
Hi there,

First of all, thank you for this amazing stuff you guys made, it's working well as expected. But, I was creating a Single-page Application using Vue.js as the front-end and Laravel as the back-end.

Currently, I am using package named `dingo/api`(https://github.com/dingo/api/) for the API things. Unfortunately, while I was there, `dingo/api` uses its own Routing (`$api = app('Dingo\Api\Routing\Router');`) and therefore, all the routes registered with it is not displayed when I run the command `php artisan route:list` but it's displayed when I run `php artisan api:routes` which is coming with the package.

So, I wonder if ziggy can work together with dingo/api so every routes that's defined there will be 'exposed' to the javascript `route()` function. It will be a great help if it could, because I personally uses dingo/api for versioning, authentication, etc that's already available out-of-box.

Thank you.

cristiancalara:
Hi @bagaskarawg 

I've had the same "issue". I've forked and implemented a quick way to send over additional routes via the routes directive. Basically I added a `$additionalRoutes` param that allows to send the routes over. I could submit a pull request if the package owners decides that this is a feature that they want to support.

For the code to get the routes in the format that Ziggy expects, I'm using:
```php
View::composer('partials.page.scripts', function ($view) {
    $api                = app('Dingo\Api\Routing\Router');
    $apiRoutes          = $api->getRoutes();
    $apiNameKeyedRoutes = collect([]);
    foreach ($apiRoutes as $version => $routesCollection) {
        $versionRoutes = collect($routesCollection->getRoutes());
        $versionNameKeyedRoutes = $versionRoutes
            ->filter(function ($route) {
                $action = $route->getAction();
                return $action && isset($action['as']) && $action['as'];
            })
            ->mapWithKeys(function ($route) use ($version) {
                $action = $route->getAction();
                $name   = $version . '.' . $action['as'];
                $data   = collect($route)->only(['uri', 'methods'])->put('domain', $route->domain());
                return [$name => $data];
            });
        $apiNameKeyedRoutes = $apiNameKeyedRoutes->merge($versionNameKeyedRoutes);
    }
    $view->with('apiZiggyRoutes', $apiNameKeyedRoutes);
});
```
And then I'm using the blade directive like: `@routes(false, $apiZiggyRoutes)`

bagaskarawg:
hi @cristiancalara,

Wow, great! Thank you for sharing with us!

[...]

PR #149, opened 2018-04-14
[WIP] Enable custom RouteCollector via config('ziggy.routeCollector').

DanielCoulbourne:
This is a kinda weird implementation of this idea and probably should use the container to resolve the classes, possibly with contracts, but it's late and I feel dumb and wanted to get this code written so I could come at it and refactor clean later.

This is all based on #107.


scheMeZa:
Could you merge from master and I'll give this a try?

PR #159, opened 2018-06-16
Add ability to specify a custom URL in the generator command

rodrigopedra:
This PR adds an `--url` option to the artisan command that generates the static route files so a user can pass a custom URL to be used by the generator.

This is useful when generating the assets for production locally, so we can specify the production url before pushing a new version.

### DIFF
--- a/src/CommandRouteGenerator.php
+++ b/src/CommandRouteGenerator.php
@@ -10,7 +10,7 @@ use Tightenco\Ziggy\RoutePayload;
 
 class CommandRouteGenerator extends Command
 {
-    protected $signature = 'ziggy:generate {path=./resources/assets/js/ziggy.js}';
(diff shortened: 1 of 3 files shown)


DanielCoulbourne:
Great PR. Would you mind writing some tests for it?

rodrigopedra:
Done! Already squashed in one commit and rewrote the commit message as per your guideline

DanielCoulbourne:
Beautiful. I'll tag it in the next day or two.

DanielCoulbourne:
Thanks for the reminder: https://github.com/tightenco/ziggy/releases/tag/v0.6.7

rodrigopedra:
Thank you very much, I was using the commit hash, but this seems better.

PR #160, opened 2018-06-19
BUGFIX Use current ziggy instance in current method

rodrigopedra:
When there is no global Ziggy instance, for example when not useing the blade directive and bundling the lib with webpack, there is not a global Ziggy, so when one calls `route().current()` it fails as this method instantiate a new `Route` instance without passing a Ziggy instance.


rodrigopedra:
Hi @DanielCoulbourne ,

 I am not sure how to write a test for it, If you can give me some directions I appreciate a lot.

DanielCoulbourne:
Thanks! This is good stuff. I need a few more days to tag a new release but this will be in there.

PR #173, opened 2018-09-18
Allow multiple route groups

adam1010:
I have a "public" route group and a "customer" route group.

This modification allows me to use the blade directive to include both route groups:

`@route(['public', 'customer'])`


DanielCoulbourne:
I will merge this eventually, but there are other PRs I want to merge first that may conflict, so I'll revisit this PR soon. Thanks for your patience.

PR #177, opened 2018-10-04
Restrict route().current() to GET request routes

### LINKED ISSUE #175: Current route returns "update" instead of "show"
### Expected Behavior
Visiting */model/{id}*, `route().current()` should return `models.show`.

### Current Behavior
Visiting */model/{id}*, `route().current()` it returns `models.update`.

### Solution
The *show* and *update* routes usually look the same, only differentiating by the call method *GET*, *PATCH*, *POST*. This means that `route().current()` sometimes picks the wrong one, as in my case. But because of obvious reasons, the current route can never be a POST/PATCH (unless using Ziggy server-side). Therefor only GET-routes should be returns by `route().current()`.

`5.7`

`0.6.8.1`

```php
Route::patch('/model/{id}', 'ModelController@show')->name('models.update');
Route::get('/model/{id}', 'ModelController@show')->name('models.show');
```
```
model.show: {uri: "model/{id}/{slug}", methods: Array(2), domain: null}
model.update: {uri: "model/{id}/{slug}", methods: Array(1), domain: null}
```

mattisbones:
This commit resolves #175 by filtering out any non-GET request routes from the array that `route().current()` checks against.


mattisbones:
I guess this could also be simplified quite a bit by putting the method check into the current filter instead:

```
const currentRoute = namedRoutes.filter(name => {
    if (this.ziggy.namedRoutes[name].methods.indexOf('GET') === -1) {
        return false;
    }

    return new Router(name, undefined, undefined, this.ziggy).matchUrl();
})[0];
```

PR #178, opened 2018-10-04
Update route generator to append routes to existing Ziggy object

### LINKED ISSUE #165: Feature Request: Appending routes
I can see a few situations where allowing multiple calls to @routes would be useful.
For example when I have a lot of shared routes for everyone and a few for just admins.

Here is an example config file:

    return [
        'whitelist' => [
            'api.*',
        ],
        'groups' => [
            'admin' => [
                'admin.*',
            ],
        ],
    ];

And blade directives:

    @routes()
    @routes('admin')

I would expect this to load both `api.*` and `admin.*` routes instead of just loading the last command used.
This will be especially helpful, when the main routes are generated using `artisan ziggy:generate` command and the admin routes are only added in admin panel for users that are already logged in.

mattisbones:
This commit resolves #165 by modifying additional calls to the `@routes()` directive to merge additional routes into the existing Ziggy object.


Robert-Hansen:
any chance this could be merged i'm working on a project where i ran into problems having multiple Ziggy objects on the window looks like this would fix it @DanielCoulbourne

DanielCoulbourne:
I'll look more closely at this tonight.

Robert-Hansen:
any update?

DanielCoulbourne:
Thanks for your patience. Merged and will tag soon.

PR #179, opened 2018-10-10
Adding wildcard support for current()

clementblanco:
I think adding the `*` wildcard support to `route().current('foo.*')` is pretty handy to check in your JavaScript code if the current named route is part of a potential group of routes. Pretty handy for toggling sub items in menus.

PS: I wrote tests which pass but it seems like 6 tests were already failing on `master`.


clementblanco:
Humm... interesting that Travis have those 6 tests green. 🤔
They were failing locally... Maybe my node/npm version locally?

clementblanco:
<3

Issue #184, opened 2018-11-05
Replace webpack with Microbundle?


bradlc:
Hey 👋

I have a couple of questions regarding the JavaScript for Ziggy:

Would you consider replacing webpack with Rollup, via [Microbundle](https://github.com/developit/microbundle)? I have done this on a branch and the UMD bundle is now 1340 bytes gzipped, down from 2161 bytes (**38% smaller**). Another benefit of this is that there are **6 fewer dependencies**. I’d be happy to submit a pull request if you’re interested.

Also, as far as I can tell Ziggy is not available on npm. Would you ever consider publishing it?

Thanks,
Brad

DanielCoulbourne:
As far as publishing on on NPM, much of the functionality of Ziggy depends on Laravel, and thus needs to be published on Composer. We COULD publish a version of it, but it would  belife  missing a lot of features.

As far as switching to roll up, I want @ankurk91 to weigh in on that as he’s our resident JS build pipeline expert.

ankurk91:
### quick dependency size test
```
cd ~
mkdir microbundle-test
cd microbundle-test
yarn init -y
yarn add microbundle --dev
du -sh node_modules
# 152 MB
```
```
cd ~
git clone https://github.com/tightenco/ziggy.git -b master --single-branch --depth=10
cd ziggy
yarn install
du -sh node_modules
#  63 MB including `mocha`
```
Current solution is half in size. 

microbundle is wrapper around rollup, i want the solution to be close metal.
If you can come up with a pure rollup [solution](https://github.com/rollup/rollup-starter-lib) that would be great. 

**Your solution should satisfy below requirements.**
* UMD build
* `babel-preset-env`; so that we can specify the browsers that Ziggy will support
* IE 11 support via `babel-plugin-transform-object-assign` polyfill
* expose `window.route` so that it works without any module bundler
* generate minified and non-minified files both

Lastly; I could have gone with `rollup` [initially](https://github.com/tightenco/ziggy/pull/87) but didn't because webpack user base is larger than rollup and it is easy to find support and help.

DanielCoulbourne:
What I'm gathering from the comments on #185 and #188 is that we can gain a much smaller minified file if we get rid of IE11 polyfills, but if we choose to keep polyfills we end up with a bundle around the same size.

Is that about right @ankurk91 and @bradlc ?

ankurk91:
[...]

PR #185, opened 2018-11-06
[WIP] Webpack to rollup - babel

ankurk91:
Hi, @DanielCoulbourne 

The PR is the follow up of https://github.com/tightenco/ziggy/issues/184

* Remove `webpack` and add `rollup`
* Upgrade babel to 7
* Source code is untouched
* The `dist` files should work in browser and module environment as before
     - Still have IE 11 support as before
* The dev dependencies size has been reduced to 23 MB (63 MB previously)
* Loose ability to use ES6 syntax in tests

@bradlc
Please check the outfile file size now. 
I am new to `rollup`, (just learned today), happy to hear recommendation.


bradlc:
Thanks for taking a look at this @ankurk91 

From my perspective the main goal of moving to Microbundle or a custom Rollup build would be to reduce the final bundle size. Unless I’m reading it wrong (I might be) it looks like the bundle produced by this setup is pretty much the same size as the current one.

In contrast, Microbundle produces a bundle that is 38% smaller than the current one. I recommended Microbundle because it has already done the hard work of setting up an optimized Rollup config, and means you only have to maintain a single dependency. I personally don’t think it matters that the total dependency size is larger.

If you are against using Microbundle then that’s cool, but there might be some Rollup nuggets in that repo that might help get your bundle size down.

ankurk91:
`microbundle` is using less popular [buble](https://github.com/Rich-Harris/buble) instead of [Babel](https://github.com/babel/babel).

`microbundle` is using [terser](https://github.com/terser-js/terser) instead of  [UglifyJS](https://github.com/mishoo/UglifyJS2)

I notice that bundle size goes to from `12` KB to `7.77` KB  when switching from Babel to buble. 
But also; it does not include the polyfills for older browsers and thats what was consuming the remaining size.

I am not confident to go with both of these. Specially `buble`.
You can purpose your solution (microbundle) in a PR and lets see what @DanielCoulbourne says.

PR #187, opened 2018-11-06
Expose dynamic parameters for current route

### LINKED ISSUE #183: Feature Request: Expose dynamic parameters for current route
### Expected Behavior
It can be useful to have all route parameters available as an object, e.g. to make asynchronous calls based on an ID. Taking an example from the README, a route like `route('events.venues.show', [1, 2])` would provide you with:

```javascript
{
    event: 1,
    venue: 2,
}
```

As for the API, I see two options:

**1. Based on the [`current()` implementation](https://github.com/tightenco/ziggy#current)**
```javascript
route().params()        // Returns `{ event: 1, venue: 2 }`
route().params('event') // Returns `1`
route().params('foo')   // Returns `undefined`
```

**2. Based on the [Vue Router implementation](https://router.vuejs.org/api/#route-object-properties)**
```javascript
route().params          // Returns `{ event: 1, venue: 2 }`
```

Both would fall back on an empty object.

### Current Behavior
None.

Is this something you'd be open to? If so, let me know your preferred implementation and I'd gladly create a PR.

osinga:
Closes #183.

### DIFF
--- a/src/js/route.js
+++ b/src/js/route.js
@@ -55,7 +55,7 @@ class Router extends String {
         return this.template.replace(
             /{([^}]+)}/gi,
             (tag, i) => {
-		 let keyName = tag.replace(/\{|\}/gi, '').replace(/\?$/, ''),
+		 let keyName = this.trimParam(tag),
                     key = this.numericParamIndices ? paramsArrayKey : keyName,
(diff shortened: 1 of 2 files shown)


ARNHOE:
Awesome!

PR #188, opened 2018-11-07
[WIP] Webpack to rollup - buble

ankurk91:
Same as https://github.com/tightenco/ziggy/pull/185, but with [buble](https://github.com/Rich-Harris/buble)

Bundle size has been reduced to `3.43` kb from `5.78` kb (minified)
Note: `buble` build does not include any polyfills,
 for example a polyfill for [Object.assign](https://babeljs.io/docs/en/babel-plugin-transform-object-assign)

@bradlc
 Is this what you wanted? 
What was your bundle size with `microbundle`?


DanielCoulbourne:
I think this is going to end up being the winner, but I want to consult with some other folks first. This week for sure.

PR #193, opened 2018-11-15
Replace webpack with Microbundle

bradlc:
@DanielCoulbourne @ankurk91 here is my version using Microbundle

**Stats (in bytes)**
```
Before:
--------
Minified: 6752
Minified, gzipped: 2350

After:
--------
Minified: 4051 (-40%)
Minified, gzipped: 1536 (-35%)
```

**Notes**

- An `extend` function has been added in place of `Object.assign` to maintain IE support
- `Object.values` has been replaced to maintain IE support
- In addition to a UMD bundle, CommonJS and ES Module bundles have been added but can be disabled (the stats above compare the old bundle to the new **UMD** bundle)


ankurk91:
Thanks. bradlc

Here are my findings

### Pure rollup with buble (without extend), [link](https://github.com/ankurk91/ziggy/blob/patch/rollup/dist/js/route.min.js)
rollup-buble


### Microbundle by bradlc 
microbundle

Not a big difference, but still; should we go with [buble](https://github.com/Rich-Harris/buble) over [babel](https://babeljs.io/) ?

bradlc:
Hey @ankurk91 

I just want to see the bundle size reduced as much as possible :) however you want to do that is fine with me!

In your version currently there is no `Object.assign` alternative so it won’t work in IE 11, unless I missed it?

Same goes for `Object.values`. In my version I have replaced it with `Object.keys(tags).map(x => tags[x])`

DanielCoulbourne:
Closes #193 (when merged)

bradlc:
Any update on this @DanielCoulbourne @ankurk91 ?

PR #198, opened 2019-01-17
Add middleware to route data

jes490:
## Use case
I have front-end project with backend proxy which proxies requests to Laravel api with Passport installed. 
I need to know how exactly should I proxy requests based on middleware of laravel routes. 
- for **client** passport middleware I need to pass *client access_token* to api routes
- for **auth:api** middleware I need to pass *completly different access_token* to api routes

With middleware route information from ziggy routes you can distinguish such routes easily.

## Changes
Added support for **middleware** config property for ziggy. 
- if **middleware** is set to true, then all route middleware will be returned. 
- if **middleware** is set to array of middlewares, then only given middleware will be returned.
- if no **middleware** is set then no changes are introduces, default behavior stays the same.

### DIFF
--- a/src/CommandRouteGenerator.php
+++ b/src/CommandRouteGenerator.php
@@ -10,7 +10,7 @@ use Tightenco\Ziggy\RoutePayload;
 
 class CommandRouteGenerator extends Command
 {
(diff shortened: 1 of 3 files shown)


jes490:
Reopen from diff branch

PR #215, opened 2019-05-15
Change public to protected and add void return type for setUp in testcases

thoresuenert:
Update TestCases to align with new function structure of Testbench/Phpunit


jakebathman:
Travis tests are failing because of an old version of `phpunit` being installed.

Checked out and tested this locally, and all tests pass (as long as I use the `phpunit` included in the project, instead of my system-wide `phpunit` binary).

jakebathman:
These changes were incorporated in other PRs or commits. Thanks for your help @thoresuenert!

PR #216, opened 2019-05-20
Allow the baseUrl to be overridden

gms8994:
We're working on setting up multi-tenant (which actually includes sub-tenants, where the sub-tenant information is in the url path). We want to ensure that paths generated by ziggy will include the sub-tenant information, therefore we want to allow the baseUrl value to be easily customized, with a fallback to `/` when not provided.

### Expected Behavior
Allows for a new configuration option `ziggy.base_url` that contains a path instead of `/`.

### Current Behavior
Currently, all baseUrl's are assumed to be `/`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ install:
   - nvm use stable
   - npm ci
 script:
-  - phpunit
+  - vendor/bin/phpunit
   - npm run build
   - npm test
 cache:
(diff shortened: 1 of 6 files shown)


gms8994:
I had to make some unrelated changes to fix tests that seemed to be broken, which is why .travis.yml and composer.json have been modified slightly.

PR #219, opened 2019-06-13
Support for Array query params, fixes #202.

### LINKED ISSUE #202: Inconsistency in query param with array value
Given a route with a query param key that has an array as a value
`route('route', ['foo' => ['a', 'b', 'c']])`

Laravel's route helper returns `/route?foo[0]=a&foo[1]=b&foo[2]=c` while ziggy returns `/route?foo=a,b,c`. When getting the param from the request object it handles it as a string and not an array.

`5.7.*`

`0.6.0`

n1c:


DanielCoulbourne:
Preferring #228 for this but thank you

PR #221, opened 2019-06-25
Update README

mariusgen:
Mention the possibility to pass an array with the groups keys


jakebathman:
Awesome, good eye for that being missing from the readme. Thanks!

PR #229, opened 2019-08-27
Modify handling of passed parameters to more closely reflect how the Laravel route method works

### LINKED ISSUE #225: Default parameters not used if number of passed parameters matches number of found parameters in route
### Current Behavior
Passing the same number of parameters to a ziggy route (as the expected number of parameters in the route) will not use default parameters, even if the parameters passed are intended to be querystring parameters.

`5.7.28`

`0.7.1`

```php
Route::prefix('{population}')->get('/api/segmentation/segments, 'SegmentsController@show')->name('api.segments');
```
_In dev tools type `Ziggy.namedRoutes` and copy the result here._

```
... SNIP ...
​
"api.segments": {…}
domain: null
methods: Array [ "GET", "HEAD" ]
uri: "{population}/api/segmentation/segments"
<prototype>: Object { … }
... SNIP ...
```

### Ziggy call in context
```javascript
route('api.segments', { hideBehavioral: this.hideBehavioral })

/// Error: Ziggy Error: 'population' key is required for route 'api.segments'
```

I noticed in https://github.com/tightenco/ziggy/blob/master/dist/js/route.js that default parameters are only used if the count of found parameters is different than the count of supplied parameters.

I would expect default parameters to be used no matter if the count matches or not, assuming an overriding value was not passed for that parameter directly.

gms8994:
Fixes #225


gms8994:
@DanielCoulbourne This doesn't pass tests because of an issue with the PHPUnit installed + PHP 7.1 - if you'd like me to remove the 7.1 Travis dependency, I can.

gms8994:
@jakebathman Can you review, if possible? This is causing us some issue in our use.

aaronhuisinga:
@DanielCoulbourne @gms8994 I believe this breaks routes with optional parameters.

Route:
```php
Route::get('/subscribers/{subscriber}/conversations/{type}/{conversation_id?}', 'ConversationsController@show');
```

Ziggy call:
```js
return route('conversations.show', {
    subscriber: 123,
    type: 'email',
     conversation_id: 1234,
});
```

This throws the exception: Ziggy Error: 'subscriber' key is required for route 'conversations.show'

One other note - the new release was tagged as v8.1 instead of v0.8.1.

gms8994:
@aaronhuisinga I'm seeing that issue as well; interestingly, it looks like if you call 

```
return route('conversations.show', {
    subscriber: 123,
    type: 'email',
     conversation_id: 1234,
}, false);
```

everything works as expected. I'll be digging in to see if I can determine the cause of the issue.

As for the tag, I've messaged DCoul as well; I'm not sure when he'll be able to fix that up.

DanielCoulbourne:
I have reverted the version. Can someone double check that for me?

I'll have time to dig deeper into all of this on FridayIn the meantime I'll roll back to v0.8.0 as the latest released version. 

If you're able to figure it out before Friday please open a PR with a test proving that @aaronhuisinga 's error case above works and I will merge and tag a release.


@gms8994 if you need to you should still be able to use `dev-master` until then.

PR #246, opened 2019-09-30
Ensure value is actual property of defaultParameters

nanaya:
If you have empty `defaultParameters`, it'll be set to `[]`.
Then if you have route with optional parameter called `filter` (`/path/{filter?}`), it'll be replaced with `function filter() { [native code] }` because that's what `[].filter` is.

This PR ensures such thing won't happen.


jakebathman:
Thanks for this fix @nanaya!

Issue #250, opened 2019-10-16
The old 0 - Zero bug is back


pmochine:
The old bug is back: https://github.com/tightenco/ziggy/issues/157

### Expected Behavior
    route('chat.inbox', 0).url()
or 

    route('chat.inbox', {offset: 0}).url()

should result to:

    https://myapp.test/api/chat/inbox/0

### Current Behavior
    route('chat.inbox', 0).url()
or 

    route('chat.inbox', {offset: 0}).url()

results to

    Uncaught Error: Ziggy Error: 'offset' key is required for route 'chat.inbox'

The link: https://myapp.test/api/chat/inbox/{offset}

The current solution is to convert the number to a string

      route('chat.inbox', '0').url()

Works fine.

### Currently installed Laravel version:
`6.0.3`

### Currently installed Ziggy version
`0.8.0`

### Example route from your Laravel Routes file.
_i.e. -_
```php
Route::prefix('/chat')->group(function () {
            Route::post('/inbox/{offset}', 'Messages\InboxController@show')->name('chat.inbox');
```
### Contents of Ziggy.namedRoutes
_In dev tools type `Ziggy.namedRoutes` and copy the result here._
chat.inbox:
domain: null
methods: ["POST"]
uri: "api/chat/inbox/{offset}"

### Ziggy call in context
_i.e. –_
```javascript
// Api client is a wrapper around Axios
console.log(axios.post(route('chat.inbox', { offset: state.inboxOffset })));
```

### EDIT1:

So now it's getting weird. If I test it:

```
it.only('Should pass param if its value is zero', function() {
        assert.equal(
            'http://myapp.dev/posts/0',
            route('posts.update', 0).url()
        );
    });
```

The test passes. But if I test on my website with ziggy 0.8.0, everything fails...

For example: `route('afterpay.update', 0).url()`

```
afterpay.update:
domain: null
methods: ["PATCH"]
uri: "processing/payment/{order}"
> Uncaught Error: Ziggy Error: 'order' key is required for route 'afterpay.update'
```

pmochine:
Starting to dig: 

If you add this test, it's going to fail: 

    it('Should pass param if its value is zero', function() {
        assert.equal(
            'http://myapp.dev/users/0',
            route('team.user.show', 0).url()
        );
    });

Results to `Error: Ziggy Error: 'id' key is required for route 'team.user.show'` as in my example

pmochine:
[...]

PR #251, opened 2019-10-18
Fixed the zero bug - Just needed to compile dist again!

pmochine:
Refers to the issue in https://github.com/tightenco/ziggy/issues/250

However, I added an extra test for this again and removed the confusing part in the constructor of the js class. 

Thank you!


jakebathman:
Hey @pmochine thanks for this! Looks like something between the `v0.8.0` release and the current `master` fixed the bug, but glad to have your test added so we can catch any regressions. Appreciate the contribution.

PR #259, opened 2019-11-27
Replaced Object.values() with .keys().map()

ryangjchandler:
Whilst using Ziggy for some AJAX on IE11, I noticed that the Object.values() method isn't being polyfilled by Babel / Webpack. Although the market share for IE11 is very small, I think it is best to use this method to add some better support for legacy browsers.

Tests pass still.


maxxscho:
I'm having the same issue. Fixed it with a polyfill on my project but it would be nice, if I just don't need the polyfill 😄

PR #267, opened 2020-01-09
Utilize afterResolving for Laravel 6.10 and up

axlon:
A fix was implemented in Laravel 6.10 making the framework no longer resolve the blade compiler on start up (laravel/framework#31009).

The following change utilizes this improvement by not calling the compiler directly if it's not yet been resolved.

### DIFF
--- a/src/ZiggyServiceProvider.php
+++ b/src/ZiggyServiceProvider.php
@@ -18,14 +18,25 @@ class ZiggyServiceProvider extends ServiceProvider
             return Macro::whitelist($this, $group);
         });
 
-        $this->app['blade.compiler']->directive('routes', function ($group) {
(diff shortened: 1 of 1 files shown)


bakerkretzmar:
Could we use [`callAfterResolving`](https://github.com/laravel/framework/pull/31009/files#diff-b8cefb9815a278c2c0aa3c91b2c3ec14R153-R161)? Something like:

```php
$this->callAfterResolving('blade.compiler', function (BladeCompiler $compiler) {
    $compiler->directive(/* ... */);
});
```

axlon:
@bakerkretzmar I don't think so, because this would break compatibility with lower versions of Laravel that do not have that feature.

bakerkretzmar:
Ahh right of course. Thanks!

PR #268, opened 2020-02-06
Fix webpack path in README

netpok:
Using
```
 ziggy: path.resolve('vendor/tightenco/ziggy/dist/js/route.js'),
```
in webpack config results in error messages containing:
```
"export 'default' (imported as '_route') was not found in 'ziggy'
```

This can be solved by using the source file instead of the already compiled one
```
ziggy: path.resolve('vendor/tightenco/ziggy/src/js/route.js'),
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,7 +248,7 @@ mix.webpackConfig({
     resolve: {
         alias: {
             ...
-            ziggy: path.resolve('vendor/tightenco/ziggy/dist/js/route.js'),
+            ziggy: path.resolve('vendor/tightenco/ziggy/src/js/route.js'),
         },
     },
 })


bakerkretzmar:
@netpok I'll merge this but I can't reproduce that error, where were you seeing it? In your terminal when building? In the browser console?

netpok:
Sorry for the late reply, I got it in console on the npm run command

related code:
```js
import route from 'ziggy';

Vue.mixin({
    methods: {
        route2: (name, params, absolute) => route(name, params, absolute),
    }
});
```

stevelacey:
@netpok I started getting this error today as I install things to try and polyfill IE11, the fix seems to work

netpok:
That will be it as I couldn't reproduce the error on a new project, but I'm pretty sure our frond developer polyfills everything possible on the old project.

@stevelacey Thanks for the info, it really bothered me what could be the reason.

PR #269, opened 2020-02-14
Update ziggy:generate path for Laravel 5.7+

Somethingideally:
Updated default ziggy:generate path


bakerkretzmar:
Thanks @Somethingideally! I believe this is technically a breaking change, in which case we'll include it in the next major release.

PR #278, opened 2020-03-22
Typescript support

benallfree:
This adds experimental support for Typescript definitions. I can't test it until #276 is accepted, so I hope it's okay if it stays open until then.

Note: If you'd prefer I maintain this separately in DefinitelyTyped, that's fine too 👍


bakerkretzmar:
We're gonna pass on pulling this in for now, but if you maintain it on DefinitelyTyped that would be awesome. Thanks!

PR #280, opened 2020-04-13
Fix handling of null optional parameter

nanaya:
This will match current laravel behavior of not throwing error when passed null for optional parameter.


jakebathman:
Pushed a rebased version of this branch on our latest `master` so GitHub Actions could run.

bakerkretzmar:
Thanks @nanaya!

PR #287, opened 2020-04-28
Allow to set nonce on blade directive

tminich:
If you need to have a restrictive content security policy that disallows unsafe inline JavaScript you need to either use a nonce or a hash of the script content. This PR allows to use a nonce with the Ziggy blade directive.
```
@routes(false, 'example-nonce')
@routes(false, csp_nonce()) // With laravel-csp
```

### DIFF
--- a/src/BladeRouteGenerator.php
+++ b/src/BladeRouteGenerator.php
@@ -27,12 +27,13 @@ class BladeRouteGenerator
         return RoutePayload::compile($this->router, $group);
     }
 
-    public function generate($group = false)
+    public function generate($group = false, $nonce=false)
     {
         $json = $this->getRoutePayload($group)->toJson();
+        $nonce = $nonce?" nonce='$nonce'":'';
 
         if (static::$generated) {
-            return $this->generateMergeJavascript($json);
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
Cool. If I understand correctly, you're basically going for this, right?

[image](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Unsafe_inline_script)

Are you currently able to use the second option, generating a hash of the script contents output by `@routes`?

tminich:
Yes, that's it.
I was considering a hash as well, but the issue with that is that the hash needs to be calculated in the blade directive and somehow communicated back to be set in the CSP header. I couldn't come up with a simple way to do that. A nonce is much easier as it can be set first and then passed to the directive.

bakerkretzmar:
Gotcha. Yeah we could generate the hash for you, but getting it into the header would still be tough because I think the Blade directive gets called late in the request. If you had to, you could use a [terminable middleware](https://laravel.com/docs/7.x/middleware#terminable-middleware) to add it to a header at the end, but the nonce is definitely easier.

PR #290, opened 2020-05-01
Fix assets being generated in current working directory

### LINKED ISSUE #282: Routes js file path is relative to current dir rather than project root
### Expected Behavior
Routes js file should be generated in the path relative to the project root.

### Current Behavior
File is generated relative to the current directory when the command is run.
Eg. my CI deploy command:
/usr/bin/php ~/releases/1/artisan ziggy:generate "./resources/js/generated/ziggy.js" -vvv
The file was generated at ~/resources/js/generated/ziggy.js rather than ~/releases/1/resources/js/generated/ziggy.js

`7.5.2`

`0.9.0`

### More info
I couldn't figure out for a while why the file wasn't being generated (but it actually was, just not where I expected it to be). I was running the command with -vvv but there's no debug output, it would be useful to show the final path that the file is generated to in this case. That would have made things much easier to figure out.

Thanks for a great package though!

bakerkretzmar:
This PR uses Laravel's `base_path()` helper to ensure that the `ziggy:generate` command always outputs assets to the correct location, even when run somewhere other than the project root.

I added a test to ensure this behaves correctly. In order for the test to work, and because Ziggy will now always use an absolute file path internally, I updated the tests to use the real filesystem instead of a virtual one. The files are generated inside Testbench's dummy Laravel app, at `vendor/orchestra/testbench-core/laravel/...`, and the new `tearDown()` method ensures they're deleted between tests. The `vfsstream` dependency was no longer needed, so I removed it.

Fixes #282.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,8 +26,7 @@
         }
     },
     "require-dev": {
-        "orchestra/testbench": "^5.0",
-        "mikey179/vfsstream": "^1.6"
+        "orchestra/testbench": "^5.0"
     },
     "extra": {
         "laravel": {
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -9,7 +9,6 @@
          processIsolation="false"
          stopOnError="false"
          stopOnFailure="false"
-         syntaxCheck="true"
          verbose="true"
 >
     <testsuites>
--- a/src/CommandRouteGenerator.php
+++ b/src/CommandRouteGenerator.php
@@ -37,7 +37,7 @@ class CommandRouteGenerator extends Command
 
         $this->makeDirectory($path);
 
-        $this->files->put($path, $generatedRoutes);
+        $this->files->put(base_path($path), $generatedRoutes);
         
         $this->info('File generated!');
     }
@@ -91,8 +91,8 @@ EOT;
 
     protected function makeDirectory($path)
     {
-        if (! $this->files->isDirectory(dirname($path))) {
-            $this->files->makeDirectory(dirname($path), 0777, true, true);
+        if (! $this->files->isDirectory(dirname(base_path($path)))) {
+            $this->files->makeDirectory(dirname(base_path($path)), 0777, true, true);
         }
         return $path;
     }
--- a/tests/Unit/CommandRouteGeneratorTest.php
+++ b/tests/Unit/CommandRouteGeneratorTest.php
@@ -3,9 +3,6 @@
 namespace Tests\Unit;
 
 use Illuminate\Support\Facades\Artisan;
-use org\bovigo\vfs\vfsStream;
-use org\bovigo\vfs\vfsStreamDirectory;
-use org\bovigo\vfs\vfsStreamWrapper;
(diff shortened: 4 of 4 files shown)


rodrigopedra:
I know this PR was for good, but it broke my setup and I spent some time trying to figure put what happened.

With previous ziggy versions I was using the `resource_path` helper in my `config/ziggy.php`.

So after the last update ziggy exported its js file to something like this:

```bash
/home/rodrigo/code/project/home/rodrigo/code/project/resources/js/ziggy.js
```

Took me a while to find the related release note. But now I have it fixed. 

Maybe it is worth to add a note about this behavior change on the readme, in case someone else were already using an absolute path like me.

Please don't take as a rant, I really like the lib and appreciate all the hard work and effort you put in it. Thanks a lot.

bakerkretzmar:
@rodrigopedra thanks a lot for letting us know! Sorry for the confusion, I'll come back and try to make a note of this in the Readme for other people who might run into it.

I'm curious about how this happened, can you share that section of your `config/ziggy.php`? I didn't think we had a config option for the output path.

rodrigopedra:
Indeed ziggy did not use that configuration You're correct, I apologize for the false reporting.

My confusion was due that for a particular project I need to export group of routes in different paths (public routes and admin panel routes).

To ease that I wrote a custom command that calls ziggy's own command underneath and for convenience I stored those paths in the same ziggy config file.

After updating it broke my frontend pipeline that called my command instead of ziggy's.

So never mind and sorry for any inconvenience. It was a total confusion on my side.

Thanks a lot for the great package and for the quick response.

For reference below is my custom command:

~~~php
<?php

namespace App\Console\Commands;

use Illuminate\Console\Command;
use Illuminate\Contracts\Config\Repository;

class GenerateZiggyRoutes extends Command
{
    protected $signature = 'app:ziggy';

    public function handle(Repository $config)
    {
        $paths = $config->get('ziggy.paths', []);

        foreach ($paths as $group => $path) {
            $this->call('ziggy:generate', [
                'path' => $path,
                '--group' => $group,
            ]);

            $this->info("[ZIGGY] {$group} routes generated");
        }
    }
}
~~~

And my config file:

~~~php
<?php

return [
    'blacklist' => ['debugbar.*', 'horizon.*'],

    'paths' => [
        'admin' => 'resources/assets/js/_admin/ziggy.js',
        'app' => 'resources/assets/js/_app/ziggy.js',
    ],

    'groups' => [
        'admin' => [
            'home',
            'admin.*',
        ],

        'app' => [
            'home',
            'app.*',
        ],
    ],
];
~~~

Maybe it would be a good idea to add some similar feature to ziggy itself. Let me know if you think this is a good idea and I can try sending a PR.

bakerkretzmar:
Thanks a lot for sharing that command and config, that's really cool. I can definitely see the use case for this—we already more or less support it via the `@routes` directive, it accepts a group parameter so you can load different groups on different pages.

Feels very related to #272 and likely wouldn't be tough to implement. If you have time to PR this please do!

PR #296, opened 2020-05-30
Fix '.' escaping in current()

davejamesmiller:


bakerkretzmar:
@davejamesmiller can you explain what this does and why it's necessary? Thanks.

davejamesmiller:
@bakerkretzmar

From the docs:

> To check whether you are currently on a specific route, you can pass the route name to current():
> route().current('events.index');
> // returns true

The `.` is not currently being escaped correctly, so `eventsXindex` would also match.

Specifically in my case, it means `hosting.*` is compiled to `^hosting..*$` instead of `hosting\..*$`, so `current('hosting.*')` returns true for the route `hosting-contacts.index` as well as `hosting.index`.

davejamesmiller:
It looks like there was originally [an attempt to escape it](https://github.com/tightenco/ziggy/pull/179/files), but it didn't work correctly because `'\.' === '.'` in JS, and then it was [auto-formatted](https://github.com/tightenco/ziggy/commit/5607f0052b0522b5e4493dffa3069620540342ff#diff-66657afd4dc58cd04bda1ac2976dcc19R170) into `.replace('.', '.')` which is more clearly a no-op.

bakerkretzmar:
Wow, `'.' === '\.'` is truly messed up....

Thanks a lot for clarifying, and for the fix! Gonna try to merge into the `0.9.x` branch so we can release it as a patch.

bakerkretzmar:
Nope I'm not, that got very messy in git very quickly 😂 will do a patch release separately.

davejamesmiller:
Thanks @bakerkretzmar 😃

PR #300, opened 2020-06-08
Change whitelist and blacklist to only and except

bakerkretzmar:
This PR renames Ziggy's `whitelist` and `blacklist` functionality to `include` and `exclude`. 

Suggestions for further tweaks to this language are welcome, I played around with `allow`/`deny` too but Ziggy isn't actually really 'allowing' or 'denying' anything, so I think `include` and `exclude` are more accurate.


mattstauffer:
I do wish we had a better word than "include", because it's often additive instead of "only". 

But I can't think of a better term. I'm all for renaming as a concept though.

bakerkretzmar:
I'm going to update this to use `only` and `except`. Among other advantages, those are the names used for similar functionality in Laravel's `Arr` and `Collection`, so they'll already be familiar to a lot of people.

PR #306, opened 2020-06-26
Remove Route facade macros

bakerkretzmar:
This PR removes Ziggy's `Route` macros, `Route::only()` and `Route::except()`.

These macros modify the global `Route` Facade, which feels feel icky and not particularly useful. More importantly, now that these methods are called `only` and `except` (they were previously `whitelist` and `blacklist`) it's not clear at all what they're actually doing or that they originate from Ziggy.


mattstauffer:
Looks good to me. If we get complaints, we can reintroduce it with a name that's scoped to Ziggy after the 1.0 release.

PR #307, opened 2020-06-26
Add support for custom scoped route model bindings

bakerkretzmar:
This PR adds support for [custom scopes in route model bindings](https://laravel.com/docs/7.x/routing#implicit-binding).

Laravel 7 added the ability to scope nested route model bindings on a specific attribute:

```php
// This route will look for a Post using the 'slug' attribute
Route::get('users/{user}/posts/{post:slug}', function (User $user, Post $post) {
    return $post;
})->name('users.posts');
```

Ziggy already supports passing an object parameter and automatically detecting an `id` key if one is present:

```js
route('users.posts', [{ id: 4, name: 'Jacob' }, 'ziggy-v1-announcement'])
// Returns http://example.com/users/4/posts/ziggy-v1-announcement
```

This PR adds support for passing an object as a parameter with a custom binding scope:

```js
route('users.posts', [{ id: 4, name: 'Jacob' }, { published: false, slug: 'ziggy-v1-announcement' }])
// Now also returns http://example.com/users/4/posts/ziggy-v1-announcement
```

Similarly to the existing behaviour with `id`, this only works if there are no conflicting route parameter names, so the example above won't necessarily work if there is another parameter explicitly called just `slug`.

### DIFF
--- a/tests/Unit/CommandRouteGeneratorTest.php
+++ b/tests/Unit/CommandRouteGeneratorTest.php
@@ -44,7 +44,7 @@ class CommandRouteGeneratorTest extends TestCase
             $this->assertFileEquals('./tests/fixtures/ziggy.js', base_path('resources/js/ziggy.js'));
         } else {
             $this->assertSame(
-                str_replace(',"bindings":[]', '', file_get_contents(__DIR__ . '/tests/fixtures/ziggy.js')),
+                str_replace(',"bindings":[]', '', file_get_contents(__DIR__ . '/../fixtures/ziggy.js')),
                 file_get_contents(base_path('resources/js/ziggy.js'))
             );
         }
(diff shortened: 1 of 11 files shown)


tests/Unit/CommandRouteGeneratorTest.php
```diff
@@ -40,7 +40,14 @@ function file_is_created_with_the_expected_structure_when_named_routes_exist()
 
         Artisan::call('ziggy:generate');
 
-        $this->assertFileEquals('./tests/fixtures/ziggy.js', base_path('resources/js/ziggy.js'));
+        if ($this->laravelVersion(7)) {
+            $this->assertFileEquals('./tests/fixtures/ziggy.js', base_path('resources/js/ziggy.js'));
```
jakebathman, line 44: TIL `assertFileEquals`

bakerkretzmar: 🤯

PR #310, opened 2020-07-25
Fix routes order to work with fallback() method

davejamesmiller:
This puts the route list into the correct order when the [`fallback()` method](https://laravel.com/docs/7.x/routing#fallback-routes) is used.

The code is taken directly from Laravel core:
https://github.com/laravel/framework/blob/29c64cb6c1d8cc712463c63fcf60c0a46fe3455e/src/Illuminate/Routing/AbstractRouteCollection.php#L77-L81

**More detailed explanation:**

If the routes are defined as:

```php
Route::fallback('NotFoundController')->name('errors.404');
Route::get('something', 'SomethingController')->name('something');
```

And you are on `/something` then Laravel returns the `something` page, but Ziggy's `route().current()` method currently returns `errors.404`, and `route().current('something')` returns `false`.

(In this trivial example you could, of course, reorder the routes - but they are actually split across several packages.)


davejamesmiller:
No problem - thank you! 😃

PR #312, opened 2020-08-01
Use microbundle instead of Webpack

bakerkretzmar:
This PR replaces Webpack with [microbundle](https://github.com/developit/microbundle), which uses Rollup under the hood. Microbundle is faster, the output is smaller, and there's way less to install and configure.

I also tweaked the outputs/targets:

- `route.js` is the default (`main`, `umd:main`, `unpkg`, and `browser`), it's a UMD build that should work basically everywhere
- `route.m.js` is an ES module (`module`)
- `route.es.js` is an ES module [specifically targeting modern browsers](https://github.com/developit/microbundle#-modern-mode-), so it's a bit smaller and faster than `route.m.js`
- **there's no CommonJS/Node build**—`route.js` should just work, bundlers or packages like `esm` can be used to import `route.m.js` if necessary, and Ziggy's target environment is the browser anyway

All tests pass on all of these builds, but I'll keep playing around with this in real apps and in the browser before it's merged.

See also [Writing a small module in 2018](https://macwright.com/2018/11/26/small-modules.html)


ankurk91:
We can disable source maps with ```--sourcemap ``` option.

bakerkretzmar:
If someone doesn't want to use the default `route.js` build, they can specify which build they want to import in one of two ways:

**Import a specific file:**

```js
import route from 'ziggy-js/dist/route.es.js';
```

**Use a Webpack alias:**

```js
// in webpack.mix.js

mix.webpackConfig({
    resolve: {
        alias: {
            ziggy: 'ziggy-js/dist/route.es.js',
        },
    },
});
```
```js
// in app.js or anywhere else

import ziggy from 'ziggy';
```

PR #313, opened 2020-08-06
Ability to manually configure base details

claudiodekker:
In situations where an app is (for example) behind a load balancer with SSL termination, Ziggy is unable to properly parse the environment details required for a correctly working front-end, which causes other front-end libraries (like Vue or Axios) to make invalid requests.

Here's an example:
image

This PR solves this by changing the way these settings are gathered, and is backwards-compatible with existing behaviour. Configurable are the following options:
- `scheme` (http, https, ssh, ftp ...)
- `host` (example.com)
- `port`(443, 80, 1234 ...)

Additionally or alternatively, a user can also provide an `url` option, which **overwrites most if not all of those configurable options**, depending on what parts of the URL were provided. For instance, if you provide `//example.com`, only the `host` will be overwritten (as only it can be parsed)

### DIFF
--- a/src/Ziggy.php
+++ b/src/Ziggy.php
@@ -210,7 +210,7 @@ class Ziggy implements JsonSerializable
 
         $this->baseUrl = Str::finish($this->baseProtocol, '://')
             . rtrim($this->baseDomain, '/')
-            . ($this->basePort ? ':' . ltrim($this->basePort) : '')
+            . ($this->basePort ? ':' . ltrim($this->basePort, ':') : '')
             . '/';
     }
 }
--- a/src/Ziggy.php
+++ b/src/Ziggy.php
@@ -201,8 +201,8 @@ class Ziggy implements JsonSerializable
         $this->basePort = $settings->get('port');
 
         $this->baseUrl = Str::finish($this->baseProtocol, '://')
-            .Str::of($this->baseDomain)->rtrim('/')
-            .($this->basePort ? Str::of($this->basePort)->prepend(':') : '')
-            .'/';
+            . Str::of($this->baseDomain)->rtrim('/')
+            . ($this->basePort ? Str::of($this->basePort)->prepend(':') : '')
+            . '/';
     }
 }
--- a/src/Ziggy.php
+++ b/src/Ziggy.php
@@ -183,13 +183,21 @@ class Ziggy implements JsonSerializable
         ];
 
         $settings = Collection::make($defaults)
+            // First, we attempt to parse our App's default using Laravel's URL helper.
             ->merge(parse_url(Str::finish($defaults['url'], '/')) ?? [])
+            // Next, we overload these defaults with any Ziggy-configured settings.
             ->merge(config()->get('ziggy', []))
+            // If the user configured a URL in Ziggy, it takes precedence over
+            // all the other user-configured settings. We'll try to parse
+            // this URL, and overload the defaults with it again.
             ->when(config()->has('ziggy.url'), function (Collection $settings) {
                 return $settings
                     ->merge(['port' => null])
                     ->merge(parse_url(Str::finish($settings->get('url'), '/')) ?? []);
             })
+            // Finally, we'll accept the given URL (if any), which takes the
+            // highest priority. We'll attempt parsing it and overwrite
+            // the defaults with it one last time.
             ->when($url, function (Collection $settings) use ($url) {
                 return $settings
                     ->merge(['port' => null])
--- a/src/Ziggy.php
+++ b/src/Ziggy.php
@@ -3,6 +3,7 @@
 namespace Tightenco\Ziggy;
 
 use Illuminate\Support\Arr;
+use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
 use JsonSerializable;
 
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
Could you explain your use case in a bit more detail?

> in situations where an app is (for example) behind a load balancer with SSL termination, Ziggy is unable to properly parse the environment details required for a correctly working front-end

I'm not sure what you mean here, or what that screenshot is showing exactly.

I'm pretty sure most of this is already possible by passing a URL to Ziggy directly, as you've done in one of your tests. With a couple small tweaks your tests all pass for me locally with Ziggy as-is, before this PR, by passing in a URL:

```php
(new Ziggy)->toArray();
// baseUrl      = http://ziggy.dev/
// baseProtocol = http
// baseDomain   = ziggy.dev
// basePort     = null

(new Ziggy(false, 'ssh://tighten.co:1337'))->toArray();
// baseUrl      = ssh://tighten.co:1337/
// baseProtocol = ssh
// baseDomain   = tighten.co
// basePort     = 1337
```

claudiodekker:
Here's a screenshot that demonstrates what my situation is a bit better:

Screenshot 2020-08-07 at 09 21 46

As you can see, the request itself is loaded over HTTPS, but because there's a load balancer in our environment that [strips the HTTPS and forwards the request as HTTP (TLS termination)](https://en.wikipedia.org/wiki/TLS_termination_proxy), Laravel receives the request as-if it were a HTTP request. 

Because Ziggy's `@routes` directive **doesn't allow for passing in an URL** (that's only possible when instantiating Ziggy manually), it uses Laravel's `url()` helper instead to guess the current request details:
https://github.com/tightenco/ziggy/blob/90dcb58f062b078a632e5cc6302fa9472fdecc5a/src/ZiggyServiceProvider.php#L28
https://github.com/tightenco/ziggy/blob/90dcb58f062b078a632e5cc6302fa9472fdecc5a/src/BladeRouteGenerator.php#L5-L11
https://github.com/tightenco/ziggy/blob/90dcb58f062b078a632e5cc6302fa9472fdecc5a/src/Ziggy.php#L18-L22

At the end of the line, this cause a HTTP response to be given by Ziggy, which then causes front-end libraries to make improper calls when using Ziggy, resulting in browser blocking the request and showing the error (screenshot) I attached in the first post.

Laravel itself also has this problem, but it has the option to 'enforce' https through the `secure_url()` helper.
I hope this clarifies things a bit better.

bakerkretzmar:
Interesting, thanks for clarifying. Is your `APP_URL` http or https?

For now, can you try adding something like this directly after `@routes`:

```html
<script>
    Ziggy.baseUrl = 'https://tradus-switchboard.test/';
    Ziggy.baseProtocol = 'https';
</script>
```

I feel like the easiest solution for this would just be to _allow_ passing a URL to `@route`, but I'll think about it some more.

claudiodekker:
Hi @bakerkretzmar, sorry for the ultra-slow response, this one slipped through!
The `APP_URL` is `https`, but the `url` helper (used in [Ziggy.php:22](https://github.com/tightenco/ziggy/blob/90dcb58f062b078a632e5cc6302fa9472fdecc5a/src/Ziggy.php#L22)) actually ignores this:

1. [The url helper itself (with input `$to = '/'`)](https://github.com/laravel/framework/blob/21520eec78945c9e156186512020bd09c3d171c6/src/Illuminate/Foundation/helpers.php#L918)
2. [`URLGenerator::to('/' ...)`](https://github.com/laravel/framework/blob/7.x/src/Illuminate/Routing/UrlGenerator.php#L189-L205)
3. [`$this->formatScheme($secure = null)`](https://github.com/laravel/framework/blob/21520eec78945c9e156186512020bd09c3d171c6/src/Illuminate/Routing/UrlGenerator.php#L297-L304)
4. [Symfony's request: `getScheme()`](https://github.com/symfony/http-foundation/blob/3081f76fa297650e0152377d7194ef043ccb4733/Request.php#L932-L934)

The workaround would indeed help, but it's a bit hacky if we're talking a long-term solution (what if Ziggy changes it's JS component?).

Thanks for looking into this again!

bakerkretzmar:
@claudiodekker okay, I see what you mean. The `url()` helper reads from the request, which I assume in your case ends up being `http` by the time it reaches Laravel?

You're right, that javascript workaround isn't great. I think the best way to tackle this is to either add a URL parameter to the `@routes` directive or a config option like `ziggy.url` to just override it globally.

I'll keep looking into this.

claudiodekker:
Yep, spot on. Looking forward to the (more permanent) solution :)

bakerkretzmar:
@claudiodekker do you think this would solve your problem?

In `Ziggy.php`:

```diff
- $this->baseUrl = Str::finish($url ?? url('/'), '/');
+ $this->baseUrl = Str::finish($url ?? config('app.url'), '/');
```

That way it'll always use `APP_URL`, which I believe in your case should ensure it's `https`. I realize this isn't nearly as configurable as the other options, but I'm wondering if they're overkill? Let me know what you think.

claudiodekker:
Hi @bakerkretzmar,

Sorry for the slow reply. Yes, I believe that'll do the trick!

bakerkretzmar:
Closed by #334. @claudiodekker thanks again for thinking this through with me!

sbc640964:
I would expect the possibility to enter a base address (domain) in route() in js

I use tenancy for laravel and want to make requests from the URL of the administration into the URL of the tenants

bakerkretzmar:
@sbc640964 you can—if the Ziggy config object contains a `location` key, Ziggy will use its value instead of the window URL.

PR #330, opened 2020-09-25
Improve internal route and parameter parsing

### LINKED ISSUE #211: Is there any way to make route().current() also match params?
### Expected Behavior
Some way to determine that the route matches, including params. For example, if I'm on any of the following pages, `route().current('backend.inventory.show')` returns true because the route name is the same for all:
* /backend/inventory/1/category/2
* /backend/inventory/2/category/2
* /backend/inventory/3/category/13
* /backend/inventory/7/category/1
* /backend/inventory/1/category/13

### Current Behavior
I'm not sure if there is already some method or argument to a method that will allow me to determine this.

`5.8.15`

`0.7.1`

```php
Route::get('/backend/inventory/{tenant}/category/{category}', [Backend\InventoryController::class, 'show'])->name('backend.inventory.show');
```
```json
"backend.inventory.show": {
	"domain": null,
	"methods": [
		"GET",
		"HEAD"
	],
	"uri": "backend/inventory/{tenant}/category/{category}"
}
```

### Ziggy call in context
I have a custom Vue directive that uses the `href` attribute to determine if the link is to the current page. The link's `href` will be a route: `<a :href="route('backend.inventory.show'. { tenant: currentTenant, category: category.id })">{{ category.name }}</a>`

```javascript
export default function (el, binding, vNode) {
    let { name } = vNode.data.attrs.href;

    if (name === undefined) {
        return;
    }

    // You can ignore this part, it works as intended for those routes that I DO want to be fuzzy matched.
    if (binding.modifiers.fuzzy) {
        const parts = name.split('.');

        if (parts.length > 1) {
            parts.splice(-1, 1, '*'); // remove last segment of route and replace with fuzzy token
            name = parts.join('.');
        }
    }

    if (route().current(name)) {
        el.classList.add('active');
    }
}
```

bakerkretzmar:
This PR rewrites most of Ziggy's JavaScript core with two main goals:

1. Take advantage of the `bindings` property added in #315 to more reliably handle complex nested parameters and route model binding.
1. Don't mutate Ziggy's own internal state while it's parsing/compiling a URL—we used to `delete` properties from `this.urlParams` as we handled them, which constrained the order they could be handled in and caused other headaches.

To accomplish this, several things are now handled quite differently internally:

- Parameters are immediately parsed into an object, with parameter names as keys and plain strings or integers as values. This makes merging in default parameter values and compiling the final URL much simpler.
- Route template 'segments' are parsed into an ordered array of objects, which makes them easier to access repeatedly and to perform merge-like operations on with their parameter values.
- There is a new `Route` class that encapsulates functionality related to handling one specific route, as opposed to some or all of them.

This PR also adds one **hot new feature**: the ability to pass a second argument to `current()` with an object containing parameters to check in addition to the route name. If that sounds familiar it's because I copied and pasted that sentence from #314, which also added that feature, but this time I didn't half-ass it and so the caveat in that PR no longer applies—now, the object passed to `current()` will be _partially_ matched against the current URL, so you can pass a subset of parameters and only the ones you pass will be checked. This PR therefore closes #314 and closes #211.

This PR **removes**:

- the publicly available `name`, `absolute`, `ziggy`, `urlBuilder`, `template`, `urlParams`, `queryParams`, and `hydrated` properties on the `Router` class
- the publicly available `normalizeParams()`, `hydrateUrl()`, `matchUrl()`, `constructQuery()`, `extractParams()`, `parse()`, and `trimParam()` methods on the `Router` class

And **deprecates**:

- `with()` in favour of just passing parameters to the `route()` function
- `check()` in favour of `has()` for [consistency with Laravel](https://laravel.com/api/8.x/Illuminate/Routing/Router.html#method_has)

And uses [microbundle's property mangling](https://github.com/developit/microbundle#mangling-properties) to restrict internal APIs.

Yes, that's a lot to remove. None of these APIs were documented, and most of them were only used or intended to be used internally. Removing them and replacing their functionality with internal properties and methods (hidden by mangling them as described above) gives us a lot more flexibility to tweak their implementation later and makes Ziggy's intended uses and what its public APIs are much clearer. Most of the functionality of the removed methods that do more than simple string parsing is still available in other ways, and most of the information contained in the removed properties is available already in the global `Ziggy` configuration object.

Todos before merging:

- [x] Update Changelog


src/js/route.js
```diff
@@ -1,267 +1,218 @@
...
+            name: segment.replace(/{|\??}/g, ''),
+            required: !/\?}$/.test(segment),
+        })) ?? [];
     }
 
-    withQuery(params) {
-        Object.assign(this.queryParams, params);
-        return this;
+    // Get whether this route's template matches the given URL
+    current(url) {
```
mattstauffer, line 32: Can we please rename this to `isCurrent`

src/js/route.js
```diff
@@ -1,267 +1,218 @@
...
-        if (name) {
-            const pattern = new RegExp(
-                '^' + name.replace('.', '\\.').replace('*', '.*') + '$',
-                'i'
-            );
-            return pattern.test(currentRoute);
+    // Get the parameter values from the current window URL, based on the given route definition
+    _dehydrate(route) {
+        let pathname = window.location.pathname
+            .replace(this._config.baseUrl.replace(/^\w*:\/\/[^/]+/, ''), '') // Remove subdirectories
```
mattstauffer, line 112: New comment, maybe: 
If this Laravel app is served from a subdirectory, trim it first

src/js/route.js
```diff
@@ -1,267 +1,218 @@
...
+            .replace(/^\/+/, '');
+
+        // Given the hydrated string, template, and delimiter, extract and return
+        // an object of parameter names and values from part of a path or URL
```
mattstauffer, line 116:
```suggestion
        // Given a URL piece, a template, and a delimiter, extract and return
        // an object of parameter names and values
```

I don't love "URL piece" but hydrated doesn't follow for a first timer to the codebase

src/js/route.js
```diff
@@ -1,267 +1,218 @@
...
+        const dehydrate = (hydrated, template = '', delimiter) => {
```
mattstauffer, line 117: Can we possibly rename `hydrated` to whatever we come up with above

src/js/route.js
```diff
@@ -1,267 +1,218 @@
...
+            const [values, segments] = [hydrated, template].map(s => s.split(delimiter));
+
+            return segments.reduce((result, current, i) => {
+                // Ignore values that are empty or don't correspond to a route segment
```
mattstauffer, line 121:
```suggestion
                // For each route template segment, if it's a parameter that's set in the
                // full URL, add its key and value to our return object
```

src/js/route.js
```diff
@@ -1,267 +1,218 @@
...
+        // Test the passed name against the current route, matching some
+        // basic wildcards, e.g. `events.*` matches `events.show`
+        const match = new RegExp(`^${name.replace('.', '\\.').replace('*', '.*')}$`).test(current);
+
+        if (!params) return match;
+
+        // If parameters were passed, check that their values match in the current window URL
+        params = this._parse(params, new Route(current, route, this._config));
+
+        return Object.entries(this._dehydrate(route))
```
mattstauffer, line 157: Possible comment:
```js
// For each passed parameter, check if it's weakly equal to the same parameter in the current URL
```

src/js/route.js
```diff
@@ -1,267 +1,218 @@
...
+    // Check whether the given named route exists
+    has(name) {
+        return Object.keys(this._config.namedRoutes).includes(name);
     }
 
-    parse() {
-        this.return = this.hydrateUrl() + this.constructQuery();
+    // Add query parameters to be appended to the hydrated URL
+    withQuery(params) {
+        this._queryParams = { ...this._queryParams, ...params };
```
mattstauffer, line 176:
```suggestion
        this._queryParams = params;
```

src/js/route.js
```diff
@@ -1,267 +1,218 @@
-import UrlBuilder from './UrlBuilder';
-import { stringify } from 'qs';
+import { parse, stringify } from 'qs';
 
-class Router extends String {
-    constructor(name, params, absolute, customZiggy = null) {
-        super();
+class Route {
```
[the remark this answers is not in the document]
mattstauffer: Can we document what each class is for at the top of the class so it's clear that Route is for internal and Router is the public object?

src/js/route.js
```diff
@@ -1,267 +1,218 @@
...
-        this.name = name;
-        this.absolute = absolute;
-        this.ziggy = customZiggy ? customZiggy : Ziggy;
-        this.urlBuilder = this.name ? new UrlBuilder(name, absolute, this.ziggy) : null;
-        this.template = this.urlBuilder ? this.urlBuilder.construct() : '';
-        this.urlParams = this.normalizeParams(params);
-        this.queryParams = {};
-        this.hydrated = '';
+    constructor(name, definition, config) {
+        Object.entries({ name, ...definition, ...config }).map(([key, value]) => this[key] = value);
```
mattstauffer, line 8: Let's instead store these as `name` and `definition` and `config` props so it's clearer where each property you're calling later comes from

src/js/route.js
```diff
@@ -1,267 +1,218 @@
...
-        return Object.assign({}, params);
+        return `${origin}/${this.uri}`.replace(/\/$/, '');
     }
 
-    with(params) {
-        this.urlParams = this.normalizeParams(params);
-        return this;
+    // Get an array of objects representing the parameters that this route accepts,
+    // e.g. [{ name: 'team', required: true }, { name: 'user', required: false }]
+    get segments() {
```
mattstauffer, line 24: If we could come up with a name that indicates "parameterizedSegementsButNotTheOtherOnes" that'd be awesome, but if not this is a good name

mattstauffer (changes requested):
Freaking fantastic work

PR #331, opened 2020-09-25
Add upgrade guide and update Readme for v1

bakerkretzmar:


README.md
```diff
@@ -1,301 +1,340 @@
-![Ziggy - Use your Laravel named routes in JavaScript](https://raw.githubusercontent.com/tighten/ziggy/master/ziggy-banner.png?version=3)
+![Ziggy - Use your Laravel named routes in Javascript](https://raw.githubusercontent.com/tighten/ziggy/master/ziggy-banner.png)
```
mattstauffer, line 1: The S is capital :)

README.md
```diff
@@ -305,28 +344,151 @@ import { Ziggy } from './ziggy';
...
+```
+
+The `/api/ziggy` endpoint above that returns Ziggy's config can be something as simple as:
+
+```php
+// routes/api.php
+
+use Tightenco\Ziggy\Ziggy;
+
+Route::get('api/ziggy', fn () => response()->json(new Ziggy));
```
mattstauffer, line 415: I wonder if it would be worth sharing how to do this above when we show how to publish the routes file?

README.md
```diff
@@ -336,6 +498,51 @@ return [
 ];
 ```
 
+#### Re-generating the routes file when your app routes change
+
+Watch your app's route files, and whenever they're saved, re-run `php artisan ziggy:generate` automatically. The example is a Laravel Mix plugin but you could do it without Mix too. Huge thanks to [Nuno Rodrigues](https://github.com/nacr) for [the idea and a sample implementation](https://github.com/tighten/ziggy/issues/321#issuecomment-689150082)!
```
mattstauffer, line 503: I think here you meant to write something like "If you're exporting your Ziggy routes as a file, you may want to watch"

mattstauffer (changes requested):
Really fantastic work here. Just a few quick notes.

PR #336, opened 2020-10-09
Return a literal string from route() when generating a URL

bakerkretzmar:
This change would make using the `route()` function a little cleaner by removing the need to ever call `.url()`, bring its behaviour closer to matching Laravel's `route()` helper, and resolve a [very](https://github.com/inertiajs/pingcrm/issues/32) [annoying](https://github.com/tighten/ziggy/pull/256) [quirk](https://github.com/laravel/jetstream/pull/316) with axios that required calling `.url()` to ensure that CSRF tokens are attached to the request.

When no arguments are passed to `route()`, it will continue to return a `Router` instance:

```js
route(); // returns a Router
route().current(); // unchanged, returns a string (a route name)
route().current('home'); // unchanged, returns a boolean
route().params; // unchanged, returns an object
route().has('posts.show'); // unchanged, returns a boolean
```

However, now, if it's passed a `name` argument, it generates the URL and returns a string directly:

```js
// before
route('home'); // returns a Router, which extends the String class, with a value of 'https://example.com'
route('home').url(); // returns 'https://example.com'

// after
route('home'); // returns 'https://example.com'
```

This PR also:

**Removes the `.with()` method**

This method was going to be deprecated in `1.0` since it isn't really necessary, parameters should always be passed as the second argument to `route()`.

**Removes the `.withQuery()` method**

Query parameters can also be passed in the second argument to `route()`, mixed in with regular route parameters.

For edge cases like where a query parameter has exactly the same name as a route parameter, this PR adds a special `_query` parameter that will be added directly to the query string:

```js
route('posts.show', {
    post: 4,
    _query: {
        post: 5,
    },
});
// 'https://example.com/posts/4?post=5'
```

**Removes the `.url()` method**

Calling `.url()` is no longer possible or necessary since `route(...)` with any arguments returns a string. To reduce repetition it was removed and its functionality was moved to the existing `.toString()` method on the Router class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Breaking changes are marked with ⚠️.
 - Use [microbundle](https://github.com/developit/microbundle) instead of Webpack to build and distribute Ziggy ([#312](https://github.com/tighten/ziggy/pull/312))
 - ⚠️ Default Ziggy's `baseUrl` to the value of the `APP_URL` environment variable instead of `url('/')` ([#334](https://github.com/tighten/ziggy/pull/334))
 - ⚠️ Allow getting the route name with `current()` when the current URL has a query string ([#330](https://github.com/tighten/ziggy/pull/330))
+- ⚠️ Return a literal string from the `route()` function when any arguments are passed to it ([#336](https://github.com/tighten/ziggy/pull/336))
 
 **Deprecated**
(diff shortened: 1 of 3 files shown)


jakebathman (approved):
Looks good, I'm excited about this one 👍

PR #337, opened 2020-10-09
Remove baseProtocol and baseDomain

### LINKED ISSUE #333: Explore whether baseProtocol config option can be removed
Ziggy only deals with Laravel named routes, which as far as I know can only be `http` and `https`. If that's the case, URL protocols probably don't need to be configurable.

See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48210#discussion_r495505368

bakerkretzmar:
This PR removes the `baseProtocol` and `baseDomain` properties on the Ziggy configuration object. `baseDomain` actually wasn't used anywhere, and `baseProtocol` was inferred from `baseUrl`, so now we just use that directly.

Closes #333.

### DIFF
--- a/src/Ziggy.php
+++ b/src/Ziggy.php
@@ -10,9 +10,7 @@ use ReflectionMethod;
 
 class Ziggy implements JsonSerializable
 {
-    protected $baseDomain;
     protected $basePort;
-    protected $baseProtocol;
     protected $baseUrl;
     protected $group;
     protected $routes;
@@ -22,12 +20,7 @@ class Ziggy implements JsonSerializable
         $this->group = $group;
 
         $this->baseUrl = rtrim($url ?? config('app.url', url('/')), '/');
-
-        tap(parse_url($this->baseUrl), function ($url) {
-            $this->baseProtocol = $url['scheme'] ?? 'http';
-            $this->baseDomain = $url['host'] ?? '';
-            $this->basePort = $url['port'] ?? null;
-        });
+        $this->basePort = parse_url($this->baseUrl)['port'] ?? null;
 
         $this->routes = $this->nameKeyedRoutes();
     }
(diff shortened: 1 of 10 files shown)


src/Ziggy.php
```diff
@@ -22,12 +20,7 @@ public function __construct(string $group = null, string $url = null)
         $this->group = $group;
 
         $this->baseUrl = rtrim($url ?? config('app.url', url('/')), '/');
-
-        tap(parse_url($this->baseUrl), function ($url) {
-            $this->baseProtocol = $url['scheme'] ?? 'http';
-            $this->baseDomain = $url['host'] ?? '';
-            $this->basePort = $url['port'] ?? null;
-        });
+        $this->basePort = parse_url($this->baseUrl)['port'] ?? null;
```
jakebathman, line 23: Just wanna make sure, same as the other PR, that this gets updated to `$this->url` in whatever order these PRs are merged. Not necessarily here, probably in #338, but don't wanna forget.

src/js/route.js
```diff
@@ -28,7 +28,7 @@ class Route {
         // If  we're building just a path there's no origin, otherwise: if this route has a
         // domain configured we construct the origin with that, if not we use the app URL
         const origin = !this.config.absolute ? '' : this.definition.domain
-            ? `${this.config.baseProtocol}://${this.definition.domain}${this.config.basePort ? `:${this.config.basePort}` : ''}`
+            ? `${this.config.baseUrl.match(/^\w+:\/\//)[0]}${this.definition.domain}${this.config.basePort ? `:${this.config.basePort}` : ''}`
```
jakebathman, line 31: Same here, `this.config.url`

jakebathman (changes requested):
Looks good. I marked two places where `baseUrl` is new in this PR just to highlight the order of PR merging with #338. Otherwise good to go 👍

src/Ziggy.php
bakerkretzmar: 👍🏻

PR #338, opened 2020-10-09
Remove "base" and other prefixes

bakerkretzmar:
This PR renames:

- `namedRoutes` → `routes`
- `defaultParameters` → `defaults`
- `baseUrl` → `url`
- `basePort` → `port`

Shouldn't have much of an effect as they're mostly used internally, and where people are using them it's a straightforward change. I left out `baseProtocol` and `baseDomain` in anticipation of them being removed in #337.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -19,8 +19,8 @@ assignees: ''
 
 <!-- Include ALL of the information below: -->
 
-- **Laravel version**: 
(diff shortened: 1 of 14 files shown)


jakebathman (changes requested):
Everything here looks good, but in #341 you added a few instances of `namedRoutes` in `tests/Unit/RouteModelBindingTest.php` that'll need to be handled as well.

That could be addressed either in this PR or that one, depending on the order of merging. I don't think I care either way.

PR #341, opened 2020-10-15
Make internal PHP functionality private

bakerkretzmar:
This PR removes some methods and makes a few others private in Ziggy's PHP API. These methods were either for internal use or provided no real functionality of their own.

**Notable changes**

- The `filter()` method on the `Ziggy` class is now 'fluent' and returns an instance of `Ziggy` instead of a collection of routes.

**Other changes**

- The `nameKeyedRoutes()` and `resolveBindings()` methods on the `Ziggy` class, which are for internal use, are now private.
- The `generate()` method on the `CommandRouteGenerator` is now private.
- The `applyFilters()` and `group()` methods on the `Ziggy` class are now private. This might seem like a significant change, but it isn't. Both of these methods returned an array of Ziggy-formatted route definitions, which isn't particularly useful on its own, but more importantly, they were both redundant—Ziggy already uses these methods internally to filter its routes, so calling `group()` or `applyFilters()` directly has the same effect as configuring or passing the same options before instantiating a Ziggy object. The easiest way to understand this is to look at the changes to the tests in this PR: all of the data returned by `applyFilters()` and `group()` was already available in the `namedRoutes` key of the array that Ziggy returns.

**Cleanup**

- Removed the `appendRouteToList()` and `isListedAs()` methods, and code referencing them. These were both unused since the `Router` Facade macro was removed in #306.
- Removed the `except()` and `only()` methods that didn't return anything and were already being called automatically when retrieving Ziggy's config. This functionality was inlined into the `applyFilters()` method.


jakebathman (approved):
Good to merge 👍

PR #362, opened 2020-11-13
Fix error calling 'current()' on an unknown route

### LINKED ISSUE #360: Error in route().current() when on 404 page
#### Description

When I'm on the 404 page and I call `route().current('blah')`, I get this exception:

```
TypeError: Invalid attempt to destructure non-iterable instance.
In order to be iterable, non-array objects must have a [Symbol.iterator]() method.
    Babel 2
    current Router.js:71
    ...
```

I tracked it down to this block:

```js
        const [current, route] = Object.entries(this._config.routes).find(
            ([_, route]) => new Route(name, route, this._config).matchesUrl(url)
        );
```

Because `Object.entries(...).find()` returns `undefined` (no match is found), the destructuring into `[current, route]` throws the error.

Instead I think it needs to be something like:

```js
        const result = Object.entries(this._config.routes).find(
            ([_, route]) => new Route(name, route, this._config).matchesUrl(url)
        );

        if (!result) return false;

        const [current, route] = result;
```

Although I'm not sure if `route().current()` with no parameters should return `false`, `null` or something else.

#### Expected behavior

`route().current('blah')` returns `false`.
`route().current()` returns `false`, `null` or something other than an error.

#### Environment

- **Laravel version**: 8.14.0
- **Ziggy version**: 1.0.1

**Related routes**:

Probably unimportant since it doesn't match any of them.

**`Ziggy.routes`**:

Probably unimportant since it doesn't match any of them.

**Ziggy call and context**:

It's buried in a mixin but basically `route().current('blah')`.

bakerkretzmar:
After this PR, instead of causing an error, calling `route().current()` on an unknown route (e.g. a route without a name) now returns `undefined`, and passing an invalid route name, for example `route().current('this-route-name-doesnt-exist')`, returns `false`.

Fixes #360.

### DIFF
--- a/src/js/Router.js
+++ b/src/js/Router.js
@@ -76,7 +76,7 @@ export default class Router extends String {
         // Find the first route that matches the current URL
         const [current, route] = Object.entries(this._config.routes).find(
             ([_, route]) => new Route(name, route, this._config).matchesUrl(url)
-        );
+        ) || [undefined, undefined];
 
         // If a name wasn't passed, return the name of the current route
         if (!name) return current;
--- a/src/js/Router.js
+++ b/src/js/Router.js
@@ -66,7 +66,7 @@ export default class Router extends String {
      *
      * @param {String} name - Route name to check.
      * @param {(String|Number|Array|Object)} params - Route parameters.
-     * @return {(Boolean|String)}
+     * @return {(Boolean|String|undefined)}
      */
     current(name, params) {
         const url = this._config.absolute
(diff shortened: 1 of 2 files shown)


davejamesmiller:
Thanks for the quick fix. 👍

dbcreator-studios:
route().current() still return undefined in latest version

bakerkretzmar:
@dbcreator-studios if you call `route().current()` on an unnamed or unknown route, returning `undefined` is the expected behaviour.

PR #371, opened 2020-11-20
Ensure Ziggy.defaults is always an object

### LINKED ISSUE #366: route() is returning a function
#### Description
Since upgrading to v1, I am getting a really weird result on one of my routes. From my Vue component, it is returning my route parameter as a function definition, eg:

`http://codinglabs.test:3000/hq/shifts/function%20shift()%20%7B%20%5Bnative%20code%5D%20%7D?id=6&employee_id=3&cycle_id=1&project_id=2&date=2020-11-04&created_at=2020-11-16T03%3A59%3A38.000000Z&updated_at=2020-11-16T03%3A59%3A38.000000Z&project=2`

`shift()` is a computed property on this component, but it returns an object or undefined.

Below are some various results I got from the console:

image

#### Expected behavior
Like the working `route('hq.projects.show')` above, I expect ziggy to determine the correct route with the object from the ID.

#### Environment

- **Laravel version**:  8.14
- **Ziggy version**:  1.0.2

**Related routes**:

```php
Route::middleware('auth')->prefix('hq')->name('hq.')->group(function () {
  // these are all fine
  Route::resource('projects', Controllers\Hq\ProjectController::class);   

  // only the store method is working with route()
  Route::resource('shifts', Controllers\Hq\ShiftController::class)->only(['store', 'update', 'destroy']);
});
```

**`Ziggy.routes`**:

```
hq.shifts.update:
bindings: {shift: "id"}
methods: (2) ["PUT", "PATCH"]
uri: "hq/shifts/{shift}"
__proto__: Object
```

**Ziggy call and context**:

```js
// Shift.vue methods
      update(project) {
        this.$inertia.put(route('hq.shifts.update', this.shift), {
          project_id: project.id,
        }, {
          preserveScroll: true
        })
      },
```

**Theory**
I figured it had to be some kind of naming conflict, so I:
- commented out my `shift()` method ❌
- renamed `shift` to `shifty` in the component ❌
- renamed the route parameter `{{shift}}` to `{{shifty}}` in the routes ✅

So I guess `shift` is getting evaluating in JavaScript somehow? I'd like to do a PR but unfortunately it is a bit beyond my js abilities.

bakerkretzmar:
This PR fixes #366 by ensuring that the JavaScript `Ziggy.defaults` property is always an object, even when it's empty.

The bug in #366 happens because previously, when `defaults` was an empty PHP array, it was encoded as an empty JavaScript array too. JavaScript arrays of course have several built-in methods, like `shift()` and `reduce()`, and so if you had a route parameter name that matched one of those methods, and you tried to access it on an array, it would return that method:

```js
const defaults = [];
const name1 = 'users';
const name2 = 'shift';

!!defaults[name1]; // false, because it evaluates to defaults['users'] which doesn't exist
!!defaults[name2]; // true, because it evaluates to defaults['shift'] which DOES exist and is a function
```

If there _are_ default parameters, `defaults` will be an object, which avoids this issue, but if there aren't, it would have been an array, so any route parameter names matching built-in JavaScript array methods would cause errors like this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     },
     "scripts": {
         "build": "microbundle --name route --format modern,es,umd --external none",
+        "watch": "npm run build watch",
         "build:npm": "microbundle --name route --format modern,es,umd",
         "test": "jest",
         "prepublishOnly": "npm run build:npm"
--- a/src/Ziggy.php
+++ b/src/Ziggy.php
@@ -128,7 +128,9 @@ class Ziggy implements JsonSerializable
      */
     public function jsonSerialize(): array
     {
-        return $this->toArray();
+        return array_merge($routes = $this->toArray(), [
+            'defaults' => (object) $routes['defaults'],
+        ]);
     }
 
     /**
--- a/tests/Unit/RouteModelBindingTest.php
+++ b/tests/Unit/RouteModelBindingTest.php
@@ -170,7 +170,7 @@ class RouteModelBindingTest extends TestCase
             $this->markTestSkipped('Requires Laravel >=7');
         }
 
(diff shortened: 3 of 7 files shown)


sbine (approved):
🏆

PR #377, opened 2020-11-26
Fix issues with incorrect `current` result for pure param routes

TomHart:
`route().current('page', {page: 'some-path'})` returns true when on `/`, this fixes that.

I discovered this in my project, where I have the routes

```
Route::get('/', [PageController::class, 'home'])->name('home');
Route::get('/{page}', [PageController::class, 'show'])->name('page');
```

And in my nav `:active=route().current('page', 'about')` was returning true when I was on the home page

Swapping `current` in `dist/index.js` for this verified the fix works in my project too
```
current(t, e) {
            const r = this.t.absolute ? window.location.host + window.location.pathname : window.location.pathname.replace(this.t.url.replace(/^\w*:\/\/[^/]+/, ""), "").replace(/^\/+/, "/"), [n, i] = Object.entries(this.t.routes).find(([e, n]) => new N(t, n, this.t).matchesUrl(r)) || [void 0, void 0];
            if (!t) return n;
            const o = new RegExp("^" + t.replace(".", "\\.").replace("*", ".*") + "$").test(n);
            if(e){
                e = this.s(e, new N(n, i, this.t))
                var z = Object.entries(this.l(i));
                if(!z.length) return false;
                return z.filter(([t]) => e.hasOwnProperty(t)).every(([t, r]) => e[t] == r)
            }
            return o
        }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ Breaking changes are marked with ⚠️.
 ## UNRELEASED
 
 **Fixed**
-- Routes made of pure params (`{page}`) no longer get matches as when on `/`
+
+- Fix bug where `route().current()` could incorrectly return `true` on URLs with no parameters ([#377](https://github.com/tighten/ziggy/pull/377))
 
 ## [v1.0.3] - 2020-11-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 Breaking changes are marked with ⚠️.
 
+## UNRELEASED
+
+**Fixed**
+- Routes made of pure params (`{page}`) no longer get matches as when on `/`
+
 ## [v1.0.3] - 2020-11-20
 
 **Fixed**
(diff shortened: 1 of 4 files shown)


bakerkretzmar:
@TomHart thanks a lot for this! In investigating it some more I realized that the example you shared will actually return `true` for _any_ route name you pass it, if you also pass it any params... I'm going to try to see if fixing that issue fixes this one too, but either way will fix asap.

bakerkretzmar:
@TomHart just want to add thanks in particular for including tests! They were super helpful for investigating this and led me to find the other bug at play here.

For my own future reference, and anyone else who ends up here—the root cause of this issue is that JavaScript's built-in `every()` Array method **returns `true` on empty arrays**. Because we're calling `every()` on the array of parameters in the URL, not the ones passed to the `current()` method, we would almost _always_ get `true` if the URL itself doesn't have parameters, because the equality check just doesn't happen at all.

TomHart:
You're welcome :). I did notice a few other errors whilst trying to investigate this, with the `route` variable being passed to `_parse`, and it throwing errors when it tries to get `.bindings` from it, but I couldn't reproduce it once I started playing in code, if I figure it out I'll make a new PR to test/fix it

bakerkretzmar:
Hopefully today but not sure, otherwise end of next week at the latest. The `bindings` errors are from passing `current()` a route name that doesn't exist, which is one of the other bugs I'm working on now 😄

TomHart:
Sounds good! Was just curious :D

Issue #375, opened 2020-11-23
TypeError: Cannot read property 'name' of undefined


pmochine:
#### Description

Ziggy fails when you add parameters when they don't exist.

Before version 1, I could just let the code be like this:

```
  axios.post(route(endpoint, this.food.id), data)
                    .then(this.saved)
                    .catch(this.failed);
```

The endpoint is dynamically set. However, if you have URLs that don't have parameters it now fails.

```
VM12 letscook:115 Uncaught (in promise) TypeError: Cannot read property 'name' of undefined
    at VM12 letscook:115
    at Array.reduce (<anonymous>)
    at R.s (VM12 letscook:115)
    at new R (VM12 letscook:115)
    at VM12 letscook:115
    at _callee$ (offerFood.js:21)
    at tryCatch (runtime.js:45)
    at Generator.invoke [as _invoke] (runtime.js:274)
    at Generator.prototype.<computed> [as next] (runtime.js:97)
    at asyncGeneratorStep (CountryList.vue:45)
```

#### Expected behavior

It should work, even if the parameter is useless in some cases.

#### Current solution

// not that nice :D

                let url = '';

                if (endpoint === 'uploadImage') {
                    url = route(endpoint);
                } else {
                    url = route(endpoint, this.food.id);
                }

                axios.post(url, data)
                    .then(this.saved)
                    .catch(this.failed);

#### Environment

- **Laravel version**:   v8.14.0
- **Ziggy version**:  v1.0.1

**Related routes**:

```php
Route::post('/upload/image', 'Api\ImageUploadController@store')->name('uploadImage');
```

**`Ziggy.routes`**:

```js
      "uploadImage": {
            "uri": "api\/upload\/image",
            "methods": ["POST"]
        },
```

bakerkretzmar:
What is `this.food` when you aren't using it? `undefined`?

pmochine:
@bakerkretzmar thanks for the fast reply. `this.food` is a vue prop und is always set with a value. So there is no `undefined`. 

To make it short. It fails even if I write: `route('uploadImage', 'some-random-value')`

bakerkretzmar:
@pmochine okay got it, thanks. I think I understand what's happening, but I'm not sure that ignoring the extra param is actually the correct behaviour here. In an object, extra parameters are appended to the query string:

```js
route(endpoint, { id: this.food.id });
// If the route has no params, this will return "http://endpoint.test/path?id=1"
```

[...]

PR #378, opened 2020-11-26
Fix bug where you can't check current with an empty param

TomHart:
Fixed a bug where you can't check if a current route with an explicitly blank param

```
Route::get('/{page?}', [PageController::class, 'show'])->name('page');
```

`route().current('page', '')` would return true when on `/about`


bakerkretzmar:
Thanks! Pretty sure this is the same issue as #375. Working on something similar now.

TomHart:
Ahh my bad, didn't realise that was a related issue!

TomHart:
Noticed an issue where it'd trigger an error with `segments[i]` being undefined, and trying to get `.name`, so protected from that

PR #386, opened 2020-12-04
Revert #334 "Use APP_URL instead of url() helper"

### LINKED ISSUE #374: Generated routes do not retain the port number
**Description**

I'm using Ziggy with on a Laravel/Vue project. I installed and configured ziggy as per the documentation, .i.e

```js
@routes
<script src="{{ mix('/js/app.js') }}" defer></script>
```

My app.js looks like:

```js
import { createApp, h } from 'vue'
 
const app = createApp({
  // ...
})

app.mixin({
  methods: { route: window.route }
})

```

Every looks good, except my routes are missing the port number.
instead of: **localhost:8000/home**, 
I have: **localhost/home**

How can I configure it to add the port number (when I'm using on my local machine)?
Thanks

**Environment**
- **Laravel**: ^8.12
- **Ziggy**: ^1.0
- **Vuejs**: 3
- **Windows 10 Pro**

### LINKED ISSUE #376: Presence of port number in URL prevents route() from working properly
#### Description

When accessing my app at `http://gss.local:3000` (ie: when developing using Browsersync) I am unable to run things like `route().current()` or `route().parms`. However, when I access the app using the standard `http://gss.local` URL these things work fine. Switching `APP_URL` between the two to include/remove the port does not seem to make any difference.

#### Expected behavior (run from console)

`> route().current()`
expected a string containing current named route to be returned, get `undefined` insread.

`> route().params`
expected array of params to be returned. Got `Uncaught TypeError: Cannot read property 'domain' of undefined  at Router.get ((index):1258) at <anonymous>:1:8` instead.

#### Environment

- **Laravel version**: 8.12
- **Ziggy version**: 0.9.4
- **Jetstream version**: 1.6
- **Inertia version**: 0.2.4

**Related routes**:

```php
Route::middleware(['auth:sanctum', 'verified'])->get(
    '/topic/{topic}/section/{section}', 
    [App\Http\Controllers\SectionController::class, 'show']
)->name('section.show');
```

**`Ziggy.routes`**:

```js
section.show:
domain: null
methods: Array(2)
0: "GET"
1: "HEAD"
length: 2
__proto__: Array(0)
uri: "topic/{topic}/section/{section}"
__proto__: Object
```

**Ziggy call and context**:

```js

route().current()
<undefined

route().params
<Uncaught TypeError: Cannot read property 'domain' of undefined
    at Router.get ((index):1258)
    at <anonymous>:1:8

```

### LINKED ISSUE #383: Regression: Subdomain Drops off in 1.0.X Branch
#### Description

Using ziggy 1.0.X (I tested this in .0 - .3) a url that is supposed to have a subdomain does not. 

**If I revert back to 0.9.4 the subdomain issues goes away**

#### Expected behavior

`route('foo') // should return foo.mysite.com`

is actually return

`route('foo') // actually returns mysite.com`

#### Environment

- **Laravel version**:  Latest 8
- **Ziggy version**: 1.0.0 - 1.0.3

**Related routes**:

All routes with a subdomain seem to have an issue. I'm not using the Spatie multi-tenant package, however, I am routing via middleware to my tenants and NOT using laravel's `domain` method in my route calls

```php
Route::get('login', Login::class)->name('login');
```

**`Ziggy.routes`**:

I use `@routes` I can copy and paste the output if you need it, but it's a lot so just let me know and I can do that, I didn't want to just dump it all in here...

bakerkretzmar:
This PR reverts #334 and switches back to using `url('/')` as the base URL, instead of the value of the `APP_URL` environment variable. That change caused subdomains and ports to stop working, which are both quite common use cases.

This PR will cause issues again in environments like those described in #313. @claudiodekker I'll work an alternative solution—sorry to break this again, but it's ended up causing more problems than it solved.

Fixes #383, fixes #376, and fixes #374.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Breaking changes are marked with ⚠️.
 
 - Fix bug where `route().current()` could incorrectly return `true` on URLs with no parameters ([#377](https://github.com/tighten/ziggy/pull/377))
 - Fix several other bugs in `route().current()` with params ([#379](https://github.com/tighten/ziggy/pull/379))
+- Revert [#334](https://github.com/tighten/ziggy/pull/334), default Ziggy's `url` back to `url('/')` instead of the `APP_URL` environment variable ([#386](https://github.com/tighten/ziggy/pull/386))
+
 
 ## [v1.0.3] - 2020-11-20
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -32,7 +32,6 @@ Ziggy `1.0` includes significant improvements and changes, most of which won't r
   - [`baseProtocol` and `baseDomain` properties removed](#user-content-base-protocol-domain-removed)
   - [`base` and other prefixes removed](#user-content-prefixes-removed)
   - [`filter()` method made fluent](#user-content-filter-fluent)
-  - [Base URL defaults to `APP_URL`](#user-content-app-url)
   - [Unused PHP methods removed](#user-content-unused-php-removed)
   - [Internal PHP methods made private](#user-content-internal-methods-private)
   - [Undocumented Javascript methods removed](#user-content-undocumented-methods-removed)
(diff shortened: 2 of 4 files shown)


rennokki:
I've been digging down thinking it's because of my S3 deployment. This needs to be merged ASAP.

jakebathman (approved):

jakebathman:
Quick review looks good 👍

PR #392, opened 2021-01-05
Small change to docs on Webpack Mix

SecondeJK:
Laravel's default webpack mix file doesn't contain an import for path, so following the docs for someone less experienced will result in a compile error for the ziggy alias import. This is a slight one line tweak in the docs for beginners.


bakerkretzmar:
Thanks!

PR #399, opened 2021-01-17
Add more config options

### LINKED ISSUE #387: Allow overriding URL in Blade directive or config
See #313, #334, and #386.

### LINKED ISSUE #388: Set output path in config
#### Description
I want to output ziggy.js elsewhere than the default resources/js/ziggy.js.

#### Suggestion
Make it possible to specify output path for `ziggy:generate` in config.

#### Alternatives
Manually specifying path on command call every time.

bakerkretzmar:
This PR adds three config options: `url`, `path`, and `absolute`. Closes #387 and closes #388.

#### `url`

The new `url` config option will override Ziggy's base URL globally, ignoring the request URL and the `APP_URL` environment variable (unless of course you set it to the value of that environment variable, like in the example). A `$url` argument was also added to the `@routes` Blade directive to allow passing this value in at runtime.

#### `path`

The new `path` config option sets the output path of `php artisan ziggy:generate` and functions just like the `--path` argument accepted by that command. Passing a `--path` argument directly will override the config option.

#### `absolute`

The new `absolute` config option sets whether Ziggy will output complete URLs by default, including the protocol and domain (as opposed to just the pathname). This functions just like the third argument to the JavaScript `route()` helper function, and can still be overridden by passing that argument at runtime.

If we think we might ever want to add this to the `@routes` Blade directive too, now is probably the time to do it, and I feel like that might be a more common need than passing a URL, so maybe it should go third and URL fourth. That's a lot of arguments though... @jakebathman thoughts?

---

This PR also adds a default config file, merges the user's with it, and makes it publishable, and adds PHP 8 syntax examples to the Readme.


maggz69:
Does this allow one to make an override the base url as such 
```php 
//app.blade.php 
...
@routes('myBaseUrl.com');
...
```

Or is it limited to the params in the ziggy.php config. I know it was mentioned in one of the issues as being a probable alternative however I could not pick whether it made it to this PR

bakerkretzmar:
@maggz69 that was added in this PR but as the third argument to the `@routes` directive, so you would use it like this:

```php
@routes(null, null, 'myBaseUrl.com')

// or on PHP 8+
@routes(url: 'myBaseUrl.com')
```

README.md
```diff
@@ -481,7 +481,23 @@ To expose multiple groups you can pass an array of group names:
...
+#### Using `@routes` with a custom URL
+
+You can also pass a custom base URL to the `@routes` Blade directive to override it regardless of the current request URL.
+
+```php
+// PHP ^8.0
+@routes(url: 'https://your-url-here.com')
+
+// PHP <=7.4
+@routes(null, null, 'https://your-url-here.com')
```
amayer5125, line 500: The URLs here should be `https://example.com` which is [reserved for documentation](https://datatracker.ietf.org/doc/html/rfc2606).

amayer5125 (changes requested):

bakerkretzmar:
After stewing on this for several months I'm still not happy with it, and I'm not convinced that we even need it at all.

I'll put together some of the smaller formatting things in a follow-up PR, but the configuration part needs more thought.

maggz69:
> After stewing on this for several months I'm still not happy with it, and I'm not convinced that we even need it at all.
> 
> I'll put together some of the smaller formatting things in a follow-up PR, but the configuration part needs more thought.

@bakerkretzmar I'd like to pitch a use case where config for routes would come in handy. In the event of _domain based tenancy_ where a parent domain could exist in the env i.e. `parent.com`, using this library in an inertia front end would ideally mean that routes passed through in the blade would inherit the parent.com url rather than the tenant url which could be `a.parent.com`.

The parent domain in this context is meant to provide a central app whereas the tenant domains allow localized access per tenant.

paulocastellano:
@bakerkretzmar i cannot use absolute routes?

bakerkretzmar:
@paulocastellano yes, you can. Absolute routes are the default, to disable them pass `false` as the third argument to Ziggy's `route()` function.

PR #407, opened 2021-02-21
Add Vue 2 & 3 Plugins

davidhemphill:
This PR adds a little cute integration with Vue 2 and 3.

Usage:
```js
// app.js

// Vue 2
import Vue from 'vue'
import { Vue2Plugin } from 'ziggy';

Vue.use(Vue2Plugin)

// Vue 3
import { createApp, h } from 'vue';
import { Vue3Plugin } from 'ziggy';
import App from './App'

createApp({
  render: () => h(App),
}).use(Vue3Plugin)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,34 +342,32 @@ route('home', undefined, undefined, Ziggy);
 
 #### Vue
 
-To use the `route()` helper in Vue components, you can add a mixin to make it available globally:
+To use the `route()` helper in Vue components, you can use the Vue plugin:
 
 ```js
 // app.js
 
-import route from 'ziggy';
-import { Ziggy } from './ziggy';
+// Vue 2
+import Vue from 'vue'
+import {Vue2Plugin } from 'ziggy';
 
-Vue.mixin({
-    methods: {
-        route: (name, params, absolute, config = Ziggy) => route(name, params, absolute, config),
-    },
-});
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
Thanks! It looks to me like these plugins both require using the `@routes` directive, is that right?

davidhemphill:
Yeah, I believe so. :-)

bakerkretzmar:
I really like this idea but I feel like people who need to customize their setup at all might still have to implement it manually so that they can pass in configuration.

For people using the `@routes` directive, this ends up being longer than just `.mixin({ methods: { route } })`, and people not using `@routes` would still have to import `route` and `Ziggy` and create the mixin manually so that they can pass the `Ziggy` object in.

It occurred to me that we could make the export a function that takes config arguments and builds the plugin object dynamically, but that's unconventional for Vue plugins.

What do you think?

davidhemphill:
I'm not married to it if you don't think it's a good idea. :-)

bakerkretzmar:
Did some more research and realized: a) in this case we can use the same plugin for Vue 2 and 3, and b) you can pass options into a Vue plugin as the second argument to `use()`, which means the plugin can stay as a plain object and the extra config argument can be completely optional.

With those little added conveniences I'm super happy with this! Thanks for your patience! I'll tag a release this afternoon.

PR #415, opened 2021-04-02
Octane Prep

taylorotwell:
This PR takes care of resetting the generator state between requests when using Laravel Octane. Otherwise, the routes will be properly generated on the first request but all subsequent requests fail.

### DIFF
--- a/src/ZiggyServiceProvider.php
+++ b/src/ZiggyServiceProvider.php
@@ -2,8 +2,10 @@
 
 namespace Tightenco\Ziggy;
 
+use Illuminate\Support\Facades\Event;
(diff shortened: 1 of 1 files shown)


DanielCoulbourne:
Taylor Otwell – Ziggy Contributor 😂 ❤️

bakerkretzmar:
🏎️ 🧨

Kocal:
What happens if Octane is not installed? 🤔

bakerkretzmar:
Nothing, because the `RequestReceived` event will never be fired.

Kocal:
Oh alright, thanks!

PR #416, opened 2021-04-04
feat: video link add

shailesh-ladumor:
This is an awesome package for route and very useful. So, I created one video tutorial of this package. please add it in the doc. so, it helps the community. Thanks for this great package for the community.


bakerkretzmar:
Hi Shailesh! We prefer to only link to our own documentation, as we've found this is the easiest way to avoid any issues with external links going down or becoming out of date. I'd suggest submitting your tutorial to Laravel News as another way to get it in front of a lot of people. Thanks a lot for the PR and for making this video!

PR #429, opened 2021-06-02
fix: simplify merge

jaulz:
I think this PR should simplify the merge and also solve the Typescript issue that appears when you export it as a ts file:
```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ "debugbar.openhandler": { uri: string; methods: string[]; }; "debugbar.clockwork": { uri: string; methods: string[]; }; "debugbar.telescope": { uri: string; methods: string[]; }; "debugbar.assets.css": { uri: string; methods: string[]; }; ... 98 more ...; "team-billing.checkOut": { ...; }; }'.
```

### DIFF
--- a/src/BladeRouteGenerator.php
+++ b/src/BladeRouteGenerator.php
@@ -40,9 +40,7 @@ HTML;
     (function () {
         const routes = {$json};
 
-        for (let name in routes) {
-            Ziggy.routes[name] = routes[name];
-        }
+        Object.assign(Ziggy.routes, routes);
     })();
 </script>
 HTML;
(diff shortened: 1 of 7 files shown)


jaulz:
@bakerkretzmar could you please maybe check this as well?

bakerkretzmar:
I'm not super familiar with TypeScript—what exactly is causing that error? And is there _any_ difference at all in the behaviour of your code and the current version?

jaulz:
It actually happens when you generate the routes file and use a `.ts` extension. Since `Ziggy` does not have a type it's inferred from the actual type and then `routes` property is an object with strict keys where we cannot assign anything directly. Using `Object.assign` we can simply merge the types without caring about the types and it will work the same way in TypeScript and in plain JavaScript.

bakerkretzmar:
Thanks again! 😄

jaulz:
Oh, sorry, I missed the other parts... thanks for merging!

PR #432, opened 2021-06-04
fix: make location configurable

jaulz:
This PR will add full SSR support.

### DIFF
--- a/src/js/Router.js
+++ b/src/js/Router.js
@@ -106,14 +106,12 @@ export default class Router extends String {
      * @return {Object}
      */
     get location() {
-        const defaultHost = typeof window === 'undefined' ? '' : window.location.host || '';
-        const defaultPathname = typeof window === 'undefined' ? '' : window.location.pathname || '';
-        const defaultSearch = typeof window === 'undefined' ? '' : window.location.search || '';
+        const { host = '', pathname = '', search = '' } = typeof window !== 'undefined' ? window.location : {};
 
         return {
-            host: this._config.location.host || defaultHost,
-            pathname: this._config.location.pathname || defaultPathname,
-            search: this._config.location.search || defaultSearch,
+            host: this._config.location?.host ?? host,
+            pathname: this._config.location?.pathname ?? pathname,
+            search: this._config.location?.search ?? search,
         };
     }
 
--- a/src/js/Router.js
+++ b/src/js/Router.js
@@ -70,8 +70,8 @@ export default class Router extends String {
      */
     current(name, params) {
         const url = this._config.absolute
-            ? this.location.host + this.location.pathname
-            : this.location.pathname.replace(this._config.url.replace(/^\w*:\/\/[^/]+/, ''), '').replace(/^\/+/, '/');
+            ? this._location.host + this._location.pathname
+            : this._location.pathname.replace(this._config.url.replace(/^\w*:\/\/[^/]+/, ''), '').replace(/^\/+/, '/');
 
         // Find the first route that matches the current URL
         const [current, route] = Object.entries(this._config.routes).find(
@@ -105,7 +105,7 @@ export default class Router extends String {
      *
      * @return {Object}
      */
-    get location() {
+    get _location() {
         const { host = '', pathname = '', search = '' } = typeof window !== 'undefined' ? window.location : {};
 
         return {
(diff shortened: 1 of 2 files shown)


jaulz:
@bakerkretzmar this is my suggestion for https://github.com/tighten/ziggy/issues/431 😊 It would be great if you could have a look!
@aarondfrancis maybe also interesting for you 😊

aarondfrancis:
Love it, this would do the trick!

aarondfrancis:
Here's a workaround for now: https://aaronfrancis.com.wip/2021/using-ziggy-with-inertia-server-side-rendering

src/js/Router.js
```diff
@@ -15,7 +15,7 @@ export default class Router extends String {
         super();
 
         this._config = config ?? Ziggy ?? globalThis?.Ziggy;
-        this._config = { ...this._config, absolute };
+        this._config = { location: {}, ...this._config, absolute };
```
bakerkretzmar, line 18: This isn't strictly necessary, right? As long as we check for null/undefined when we access it below with something like `this._config.location?.host`?

aarondfrancis:
> Thanks a ton for this PR @jaulz, and for your post @aarondfrancis! This looks great to me, I'll wait for you to take a look at my questions and I'm just going to push up some formatting tweaks and then it should be good to go.
> 
> Am I right that after this PR, using Ziggy with Inertia and SSR would look something like this?
> 
> ```js
> import route from 'ziggy';
> 
> export default async function (event) {
>     return await createInertiaApp({
>         // ssr stuff...
>         setup({app, props, plugin}) {
> 
>             const Ziggy = { ...event.props.ziggy, location: new URL(event.props.ziggy.url) };
> 
>             return createSSRApp({
>                 render: () => h(app, props),
>             }).mixin({
>               methods: {
>                   route: (name, params, absolute, config = Ziggy) => route(name, params, absolute, config),
>               },
>           })
>         },
>     });
> }
> ```

Yes, I think so! And it's beautiful 😍 

Thanks @bakerkretzmar!

jaulz:
@bakerkretzmar thanks for cross checking and I incorporated your feedback 😊

jaulz:
@bakerkretzmar do you think we could merge it in this state? Otherwise I am happy to rework it.

bakerkretzmar:
Thanks a ton @jaulz!

yashiroiori:
> Thanks a ton for this PR @jaulz, and for your post @aarondfrancis! This looks great to me, I'll wait for you to take a look at my questions and I'm just going to push up some formatting tweaks and then it should be good to go.
> 
> Am I right that after this PR, using Ziggy with Inertia and SSR would look something like this?
> 
> ```js
> import route from 'ziggy';
> 
> export default async function (event) {
>     return await createInertiaApp({
>         // ssr stuff...
>         setup({app, props, plugin}) {
> 
>             const Ziggy = { ...event.props.ziggy, location: new URL(event.props.ziggy.url) };
> 
>             return createSSRApp({
>                 render: () => h(app, props),
>             }).mixin({
>               methods: {
>                   route: (name, params, absolute, config = Ziggy) => route(name, params, absolute, config),
>               },
>           })
>         },
>     });
> }
> ```

From where import "createSSRApp", I'm using inertiajs with react

bakerkretzmar:
I think SSR is still in early access for Inertia sponsors only: https://inertiajs.com/server-side-rendering

yashiroiori:
> I think SSR is still in early access for Inertia sponsors only: https://inertiajs.com/server-side-rendering

Yes, but in examples are in this page from where import createSSRApp??

bakerkretzmar:
I would guess something like `import { createSSRApp } from '@inertiajs/inertia-vue3'` but I'm not in early access so I'm not sure 😅

heychazza:
Hey guys,

> Thanks a ton for this PR @jaulz, and for your post @aarondfrancis! This looks great to me, I'll wait for you to take a look at my questions and I'm just going to push up some formatting tweaks and then it should be good to go.
> 
> Am I right that after this PR, using Ziggy with Inertia and SSR would look something like this?
> 
> ```js
> import route from 'ziggy';
> 
> export default async function (event) {
>     return await createInertiaApp({
>         // ssr stuff...
>         setup({app, props, plugin}) {
> 
>             const Ziggy = { ...event.props.ziggy, location: new URL(event.props.ziggy.url) };
> 
>             return createSSRApp({
>                 render: () => h(app, props),
>             }).mixin({
>               methods: {
>                   route: (name, params, absolute, config = Ziggy) => route(name, params, absolute, config),
>               },
>           })
>         },
>     });
> }
> ```

trying this, I am told event doesn't exist, has anything changed since for configuring SSR?

anburocky3:
What is the equalent of using React?

Issue #358, opened 2020-11-12
Class Laravel\Telescope\Contracts\EntriesRepository does not exist 


jroski:
#### Description

When using this package in conjunction with https://github.com/barryvdh/laravel-debugbar, Calling `@routes` blade directive or the artisan command triggers the error:

`Class Laravel\Telescope\Contracts\EntriesRepository does not exist`

#### Environment

- **Laravel version**: 7.x
- **Ziggy version**: 1.0

**Related routes**:

n/a

bakerkretzmar:
Double-check your `composer.json` to make sure you have Telescope installed correctly in the environment you're seeing these errors in. Do you get similar errors running other `artisan` commands (not provided by Ziggy)?

Closing because this doesn't sound like an issue with Ziggy.

jroski:
I do not have Telescope installed. I believe the error is thrown because a route is defined in the debugbar package for Telescope even if it isn't installed. I've located the specific route in the debugbar package:

`
$router->get('telescope/{id}', [
    ' uses' => 'TelescopeController@show',
     'as' => 'debugbar.telescope',
]);
`
The `TelescopeController@show` method has a dependency of `Laravel\Telescope\Contracts\EntriesRepository`, which doesn't exist unless Telescope is installed.

jroski:
Downgrading to 0.9.4 fixes the issue.

bakerkretzmar:
@jroski thanks, I'll work on this. Debugbar registers the Telescope route even if Telescope isn't installed, so when Ziggy tries to reflect its parameter typehints, they don't exist.

bakerkretzmar:
@jroski I actually can't reproduce this with Laravel 7 or 8.

Ziggy only tries to reflect/instantiate route parameters that implement the `UrlRoutable` contract, with Telescope's `EntriesRepository` doesn't, so I'm not sure what's happening here.

Is there additional error output you can provide? Can you share what specific version of Laravel and Debugbar you're on?

Thanks.

jroski:
Laravel version: 7.10.2
Debugbar version: 3.5.1

Stacktrace:
#0 \\vendor\\laravel\\framework\\src\\Illuminate\\Routing\\RouteSignatureParameters.php(25): ReflectionParameter->getClass()
#1 [internal function]: Illuminate\\Routing\\RouteSignatureParameters::Illuminate\\Routing\\{closure}(Object(ReflectionParameter))
[...]

PR #438, opened 2021-06-30
Fix for laravel versions without reflection api support (resolves #437)

### LINKED ISSUE #437: Class Laravel\Telescope\Contracts\EntriesRepository does not exist (#358)
### Ziggy version

v1.2.0

### Laravel version

v6.0.4

### Description

This is a duplicate of issue #358 :

>When using this package in conjunction with https://github.com/barryvdh/laravel-debugbar, Calling @routes blade directive or the artisan command triggers the error:
`Class Laravel\Telescope\Contracts\EntriesRepository does not exist`

I appreciate the previous issue was closed, as the root cause was fixed in [Laravel 6.18.17](https://github.com/laravel/framework/releases/tag/v6.18.17), and upgrading resolves the issue.

Unfortunately upgrading Laravel isn't an option for us, so I've created a solution to make Ziggy compatible for those using versions <6.18.17 and barryvdh/laravel-debugbar.

### Ziggy call and context

```js
N/A
```

### Ziggy configuration

```js
N/A
```

### Route definition

```php
N/A
```

alys-dext:
This PR resolves the issue originally raised in #358. I appreciate the root cause is not with Ziggy, but I've created a solution that allows backwards compatibility for those using Laravel <6.18.17 with [barryvdh/laravel-debugbar](https://github.com/barryvdh/laravel-debugbar).

### Context

The root issue was resolved in [Laravel 6.18.17](https://github.com/laravel/framework/releases/tag/v6.18.17):
- Support PHP 8's reflection API ([#33039](https://github.com/laravel/framework/pull/33039))

For those using earlier versions, I've resolved the error by bringing forward the point at which any excluded route names are filtered out. This means that excluding debugbar via the config...
```
// config/ziggy.php

return [
    'except' => ['debugbar.*'],
];
```

...now resolves the error, as previously the error was encountered before the filtering took place.

And now I can successfully generate a ziggy file 😄 

```
File generated!
```

### New Unit Tests

```
 ✔ IsExcluded can identify excluded route via except
 ✔ IsExcluded can identify excluded route via only
 ✔ IsExcluded can identify included route via only
 ✔ IsExcluded can identify included route via groups
 ✔ IsExcluded can identify included route via groups when only is also set
 ✔ IsExcluded can identify route in both excluded and groups as included
 ✔ IsExcluded can ignore excluded route when both except and only are set
 ✔ IsExcluded can ignore route when no config set
```

### DIFF
--- a/src/Ziggy.php
+++ b/src/Ziggy.php
@@ -89,7 +89,7 @@ class Ziggy implements JsonSerializable
     {
         $isExcluded = false;
 
-        // return unfiltered routes if user set both config options.
+        // return unfiltered routes if user set both config options
         if (config()->has('ziggy.except') && config()->has('ziggy.only')) {
             return $isExcluded;
         }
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
Thanks a lot for this PR and for including tests! My preference would be for this to be fixed upstream, ideally in the `laravel-debugbar` package, so I've submitted a PR there (barryvdh/laravel-debugbar#1203) to avoid registering the Telescope-related route causing this issue unless Telescope is actually installed.

Either way though, I really like your idea of not touching excluded routes at all.

alys-dext:
Thanks @bakerkretzmar ! That's certainly a lot simpler 😄

PR #451, opened 2021-08-10
Treat all parameters not matching a named route segment as query params

### LINKED ISSUE #450: Missing route model binding key on query filters
### Ziggy version

v1.3.5

### Laravel version

v8.53.1

### Description

Jumped from v0.8.1, so I realize this might have been a known/intentional breaking change, but I'm seeing unexpected behavior on an `index` route. Also v possible this is just user-error/not accounting for a change, so please steer me toward docs if I've missed something obvious. Getting the following error on a route that used to work:

`"Error: Ziggy error: object passed as 'filter' parameter is missing route model binding key 'undefined'."`

On Ziggy v0.8.1, this did what we expected, passing any applied filters via a nested `filter` object which were then "converted" to query parameters (eg, `/people?filter[name]=Dwyer`). This bit of functionality leans on Spatie's Query Builder, in case that's helpful context. Now we're getting the error above.

TIA for any suggestions/thoughts you all might have!

### Ziggy call and context

```js
Running the following:
this.route('admin.people.index', {filter: {name: 'Dwyer'}});
```

### Ziggy configuration

```js
{
    "url": "http://admin.cut.test",
    "port": null,
    "defaults": {},
    "routes": {
        "admin.people.index": {
            "uri": "people",
            "methods": [
                "GET",
                "HEAD"
            ],
            "domain": "admin.cut.test"
        }
    }
}
```

### Route definition

```php
Route::get('/people', 'PersonController@index')->name('people.index');
```

bakerkretzmar:
This PR assumes that any object keys/values passed into `route()` as parameters that do not match any named route segments are treated as query parameters.

**Before**

```js
// route 'home' at `/home`

route('home', ['foo']); // `/home?foo=`
route('home', { foo: 'bar' }); // `/home?foo=bar`
route('home', { _query: { foo: { bar: 'baz' } } }); // `/home?foo[bar]=baz`

route('home', { foo: { bar: 'baz' } }); // error
```

That last example causes an error because Ziggy sees that the value of the `foo` parameter passed in is an object, so it expects either an `id` key in that object or a key matching the binding for the `foo` route parameter—but there is no `foo` route parameter.

**After**

```js
// route 'home' at `/home`

route('home', ['foo']); // `/home?foo=`
route('home', { foo: 'bar' }); // `/home?foo=bar`
route('home', { _query: { foo: { bar: 'baz' } } }); // `/home?foo[bar]=baz`

route('home', { foo: { bar: 'baz' } }); // `/home?foo[bar]=baz`
```

Closes #450.

### DIFF
--- a/src/js/Router.js
+++ b/src/js/Router.js
@@ -178,7 +178,7 @@ export default class Router extends String {
 
         return {
             ...this._defaults(route),
-            ...this._substituteBindings(params, route.bindings),
+            ...this._substituteBindings(params, route),
         };
     }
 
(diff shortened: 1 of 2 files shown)


caseydwyer:
Any chance this might be merged in the next couple of days? I know you all are busy people, so no pressure—just trying to decide if I need to roll back to an earlier version for an event this weekend. Thanks for your work, Ziggy is a great little tool!

bakerkretzmar:
@caseydwyer will do 👍🏻 I think I actually meant to tag you in this to take a look and see if it'll solve your issue, but completely forgot. Looking at it again thought I'm pretty sure it's good to go.

caseydwyer:
Thanks @bakerkretzmar! Probably a dumb question...but should the changes to `src/js/Router.js` be available within `/dist/index.js` or one of the other `dist` files? ie, is an NPM build necessary for this PR? Again, pardon my ignorance. Ha

PR #462, opened 2021-09-22
fix: catch undefined global Ziggy

jaulz:
In case Ziggy is not defined at all as a variable, the current code dumps because `Ziggy is not defined`.


bakerkretzmar:
Yikes, can't believe I never noticed this. Thanks!

PR #466, opened 2021-09-26
Fix name transformation in current()

djfhe:
Fixes dot and wildcard replace to match all occurrences of , and * .

Without the fix `current('event.venue.index')` would match `event.venue-index` since only the first dot gets escaped.
Other Example: `current('event.venue.*')` would match `event.venue.index` and `event.venue-soon.index`.

This PR should fix that.

Tests are also provided.


bakerkretzmar (approved):
Thanks @djfhe! Good catch, and really appreciate the tests 💪🏻

PR #473, opened 2021-10-08
Add `useRoute()` React hook

### LINKED ISSUE #327: Mention React usage in Readme
### Description

I struggled a lot with the usage in a React Application.
The problem was that i don't want to put the `@routes` blade directive in the header of my application because everyone can see all routes directly if the source code is inspected.

Due to this i wanted to go with option two to generate the Ziggy file and use it as custom Ziggy in the route() function.
In the Readme i could only find the Vue solution to use a mixin for this purpose but nothing related to React.

### Suggestion
I wrote a very small helper that did the trick for me.
Maybe this one can also be mentioned in the Readme to save some time i someone has the same issue.
I called the file ziggyroutes and imported it in my application in the view where i need it.
Additionally the helper methods are mentioned in the function like current(), check() and url().

```
import route from '../../assets/ziggyroute';

console.log(route('user.login')); 
```

```
import route from 'ziggy';

export default function (name, params, absolute) {
    var self = {name: name, params: params, absolute: absolute};

    var routeshelper = {};
    
    routeshelper.current = (name) => {
        return route(undefined, undefined, undefined, Ziggy).current(name);
    }
    routeshelper.check = (name) => {
        return route(undefined, undefined, undefined, Ziggy).check(name);
    }
    routeshelper.url = () => {
        return route(self.name, undefined, undefined, Ziggy).url();
    }
    
    return route(name, params, absolute, Ziggy);
  }
```

If there are any improvements for the small helper please feel free to change it.

bakerkretzmar:
Adds a `useRoute()` React hook. If the `Ziggy` config object is available globally, usage is as simple as:

```jsx
import { useRoute } from 'ziggy-js';

export default () => {
    const route = useRoute();

    return (
        <a href={route('home')}>Home</a>
    );
};
```

Closes #327.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,22 +387,46 @@ Now you can use `route()` anywhere in your Vue components and templates, like so
 
 #### React
 
-To use Ziggy with React, start by importing the `route()` function and your Ziggy config. Because the Ziggy config object is not available globally in this setup, you'll have to pass it to the `route()` function manually:
+Ziggy includes a `useRoute()` hook to make it easy to use the `route()` helper in your React app:
+
+```jsx
+import React from 'react';
+import { useRoute } from 'ziggy-js';
+import { Ziggy } from './ziggy';
+
+export default function PostsLink() {
+    const route = useRoute(Ziggy);
+
+    return (
+        <a href={route('posts.index')}>Posts</a>
+    );
+}
+```
+
+If you make the `Ziggy` config object available globally, you can use the `useRoute()` hook without importing and passing Ziggy's configuration to it every time:
 
 ```js
 // app.js
-
-import route from 'ziggy';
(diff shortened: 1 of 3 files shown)


FelipeVa:
Is there a way you can use this with typescript?

deevus:
See here https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html

Since this package uses JSDoc it would be possible to generate definition files for the project.

mro95 (approved):
Very simple solution :)

CamKem:
Is there something similar for Vue?

bakerkretzmar:
@CamKem https://github.com/tighten/ziggy#vue

PR #475, opened 2021-10-26
fix: prefer exact domain definitions

jaulz:
This PR will sort the routes by the occurence of the template segment position in the domain. Hence, wild card domains are at the end of the list and will not be preferred over exact domains.

### DIFF
--- a/src/js/Router.js
+++ b/src/js/Router.js
@@ -17,12 +17,17 @@ export default class Router extends String {
         this._config = config ?? (typeof Ziggy !== 'undefined' ? Ziggy : globalThis?.Ziggy);
         this._config = { ...this._config, absolute };
 
+        // Sort routes initially and prefer exact domain definitions
+        this._config.routes = new Map(Object.entries(this._config.routes).sort(([a, routeA], [b, routeB]) => {
+            return (routeA.domain || '').indexOf('{') - (routeB.domain || '').indexOf('{');
+        }))
+
         if (name) {
-            if (!this._config.routes[name]) {
+            if (!this.has(name)) {
                 throw new Error(`Ziggy error: route '${name}' is not in the route list.`);
             }
 
-            this._route = new Route(name, this._config.routes[name], this._config);
+            this._route = new Route(name, this._config.routes.get(name), this._config);
             this._params = this._parse(params);
         }
     }
@@ -74,7 +79,7 @@ export default class Router extends String {
             : this._location().pathname.replace(this._config.url.replace(/^\w*:\/\/[^/]+/, ''), '').replace(/^\/+/, '/');
 
         // Find the first route that matches the current URL
-        const [current, route] = Object.entries(this._config.routes).find(
+        const [current, route] = Array.from(this._config.routes.entries()).find(
             ([_, route]) => new Route(name, route, this._config).matchesUrl(url)
         ) || [undefined, undefined];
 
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
Can you share an example of two of your route definitions, one with the exact domain and one with the wildcard? What order are they defined in in your app? And how does Laravel's PHP `route()` helper behave in this situation?

I'm very hesitant to do any sorting, in general we want to keep the routes in the same order that Laravel has them in.

jaulz:
Sorry for the delay but now I had the chance to check it:
image

image

The output in the view is simply `{{\Request::route()->getName()}}` and it's both the same controller method but just with a different route: the former having a domain (`parse_url(config('app.url'), PHP_URL_HOST)`) and the latter having a wildcard (`{domain}`).

Though, having a second thought... maybe it makes more sense to filter out the domains that do not match?

bakerkretzmar:
@jaulz in that example, what does Ziggy display on each of those pages?

jaulz:
@bakerkretzmar that depends on the order of how the routes are set up. If the route is with the dynamic domain is defined first then it looks like this:
image

image

If the route with the dynamic domain is defined last then it looks like this:
image

image

P.S.: sorry, I missed the URLs above but I hope it's clear that the second screen uses a wildcard domain.

bakerkretzmar:
Just want to be absolutely sure I understand—is this accurate?

**Wildcard `{domain}` first:**

```php
Route::domain('{domain}')->name('external:')->group(...);
Route::domain('app.com')->group(...);
```

URL | Laravel’s `Request::route()->getName()` | Ziggy’s `route().current()`
:--- |:--- |:---
https://app.com | `home.index` | **`external:home.index` (incorrect)**
https://custom.com | `external:home.index` | `external:home.index`

**Wildcard `{domain}` last:**

```php
Route::domain('app.com')->group(...);
Route::domain('{domain}')->name('external:')->group(...);
```

URL | Laravel’s `Request::route()->getName()` | Ziggy’s `route().current()`
:--- |:--- |:---
https://app.com | `home.index` | `home.index`
https://custom.com | `external:home.index` | `external:home.index`

---

Looking at that kind of makes me wonder... can you just define the wildcard custom domain last in your routes file?

jaulz:
Yep, defining the custom domain last works but it makes it a bit prone to errors so I would prefer something that is consistent to Laravel's behaviour 😊

bakerkretzmar:
> but it makes it a bit prone to errors

Can you elaborate a bit on this? What difference does it make on the Laravel side to just define the custom domain last?

bakerkretzmar:
Also @jaulz are you caching your routes? If so, is the output above with your routes cached or uncached?

bakerkretzmar:
Okay, after digging into Laravel and Symfony's routing components some more I'm almost positive this behaviour is coming from Symfony's compiled route collection. When routes are cached, exact domain matches (or routes without a domain defined) will take priority over wildcards, but when routes are not cached, routes will match in the order they're defined and registered in your app. I _think_.

Caching routes in production is extremely common (and often recommended), so we should probably handle this somehow, but I'm not sure how we would make it consistent with Laravel since Ziggy doesn't know or care if the app's routes are cached...

jaulz:
@bakerkretzmar sorry for the late response. So in summary this sounds more like an issue that should be reported to Symfony? Anyway, the "manual" fix is to define wildcards last so that's at least something we can do from developer side 😊

bakerkretzmar:
Yeah I'm not even certain that it's really a bug, I don't fully understand how the Symfony routing stuff works. If I can figure it out and it is a bug, I'll definitely submit it upstream. For now though, yes, the fix is to do it manually by ordering wildcards last. Thanks!

PR #482, opened 2021-12-03
fix: exclude internal parameters

jaulz:
As mentioned in https://github.com/tighten/ziggy/issues/480 this PR will exclude internal parameters. "Internal parameters" are parameters that are defined in the method signature but are not part of the path (i.e. and thus not needed on the client).

@bakerkretzmar I hope that the code makes my point clearer 😊  please let me know if any changes are needed.


bakerkretzmar:
This is really great, I didn't think it would be that simple to filter these kinds of parameters out. Thanks!

PR #501, opened 2022-01-15
Handle enums in route model bindings

### LINKED ISSUE #499: Support enums introduced in PHP 8.1
#### Description

In Laravel, there is a way to resolve route binding to any object (not only Eloquent model), as long as the class implements the `UrlRoutable` contract. It can be for [example](https://gist.github.com/andrzejkupczyk/8436f9cc321d5d7bafd04a3f4b753720) an [enum](https://wiki.php.net/rfc/enumerations), although currently this is not supported.

As enum may not include `__construct`, Laravel is not able to resolve it from the container using the `app()` helper function in `Ziggy.php`:
```php
// Avoid booting this model if it doesn't override the default route key name
$bindings[$parameter->getName()] = $override ? app($model)->getRouteKeyName() : 'id';
```

#### Suggestion

IMO it would be great if different resolvers could be used (per parameter maybe? based on type or interfaces it implements?). Default resolver - for backward compatibility - would be still using the `app()` helper function. This could be configured in the Ziggy's configuration file.

#### Alternatives

The simplest solution would be to pass any useful data as parameters, for example:
```php
$bindings[$parameter->getName()] = $override
    ? app($model, ['reflection' => new ReflectionClass($model)])->getRouteKeyName()
    : 'id';
```

This way it would be possible to customize registration of a binding with the container.

andrzejkupczyk:
Closes #499.

Support resolving route key names for route parameters using any [`UrlRoutable`](https://laravel.com/api/8.x/Illuminate/Contracts/Routing/UrlRoutable.html) implementation, e.g. [enum](https://wiki.php.net/rfc/enumerations).

Current implementation causes an error using the `@routes` Blade directive, because enum may not include `__construct` so Laravel is not able to resolve it automatically from the container using the `app()` helper function in `Ziggy.php`.

The proposed changes are not perfect, but do not require PHP 8.1 and are fully backward compatible 🤷🏻
I am open to any suggestions.

### DIFF
--- a/src/Ziggy.php
+++ b/src/Ziggy.php
@@ -12,6 +12,8 @@ use ReflectionMethod;
 
 class Ziggy implements JsonSerializable
 {
+    protected const DEFAULT_ROUTE_KEY_NAME = 'id';
+
     protected static $cache;
 
     protected $url;
@@ -177,12 +179,25 @@ class Ziggy implements JsonSerializable
                     && $model === (new ReflectionMethod($model, 'getRouteKeyName'))->class;
 
                 // Avoid booting this model if it doesn't override the default route key name
-                $bindings[$parameter->getName()] = $override ? app($model)->getRouteKeyName() : 'id';
+                $bindings[$parameter->getName()] = $override
+                    ? $this->determineRouteKeyName($model)
+                    : self::DEFAULT_ROUTE_KEY_NAME;
             }
 
             $routes[$name] = $scopedBindings ? array_merge($bindings, $route->bindingFields()) : $bindings;
         }
 
         return $routes;
+    }
+
+    private function determineRouteKeyName(string $model): string
+    {
+        if (function_exists('enum_exists') && enum_exists($model)) {
+            return !empty($model::cases())
+                ? $model::cases()[0]->getRouteKeyName()
+                : self::DEFAULT_ROUTE_KEY_NAME;
+        }
+
+        return app($model)->getRouteKeyName();
     }
 }
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -2,7 +2,8 @@
 <phpunit bootstrap="vendor/autoload.php" colors="true">
     <testsuites>
         <testsuite name="Ziggy Tests">
-            <directory suffix="Test.php">./tests</directory>
+            <directory suffix="Test.php">./tests/Unit</directory>
+            <directory suffix="Test.php" phpVersion="8.1" phpVersionOperator=">=">./tests/Unit81</directory>
         </testsuite>
     </testsuites>
     <php>
--- a/tests/Unit81/RouteModelBindingTest.php
+++ b/tests/Unit81/RouteModelBindingTest.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace Tests\Unit81;
+
+use Illuminate\Contracts\Routing\UrlRoutable;
+use Tests\TestCase;
+use Tightenco\Ziggy\Ziggy;
+
+class RouteModelBindingTest extends TestCase
+{
+    /** @test */
+    public function can_handle_enums_in_route_model_bindings()
+    {
+        app('router')->get('sections/{section}', function (Section $section) {
+            return '';
+        })->name('sections');
+        app('router')->getRoutes()->refreshNameLookups();
+
+        $this->assertSame([
+            'uri' => 'sections/{section}',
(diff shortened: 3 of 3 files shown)


src/Ziggy.php
```diff
@@ -177,12 +177,23 @@ private function resolveBindings(array $routes): array
...
         }
 
         return $routes;
     }
+
+    private function determineRouteKeyName(string $model): string
+    {
+        if (config()->has("ziggy.route_keys.$model")) {
+            return config("ziggy.route_keys.$model");
+        }
```
bakerkretzmar, line 195: What's an example of a `$model` you would put here?

bakerkretzmar:
Thanks for the PR! If you can add tests that would be great, otherwise I'll come back to this and try to. I like the idea of supporting enums but I'm not sure about this implementation.

src/Ziggy.php
andrzejkupczyk: Outdated diff.

andrzejkupczyk:
Okay, @bakerkretzmar, I reworked the previous, configuration-based solution.

Now, if the route model binding represents an enumeration, Ziggy uses the first available enum case to get the name of the route key. Unfortunately, I haven't found any other way to do it using reflection or anything like that.

bakerkretzmar:
@andrzejkupczyk thanks a lot. I've looked over this several times and I think there might be a misunderstanding here about what exactly the route key name bindings are doing. I'm not very familiar with enums yet so bear with me.

The route model key binding feature we added to Ziggy was for Eloquent models that use an attribute other than their `id` as their identifier in routes. Specifically, it was so that the frontend would know about these bindings, so you could pass a whole object into the JavaScript `route()` helper as one parameter and Ziggy would know which key to use as the actual parameter value.

For example:

```php
// Product model
class Product extends Model
{
    public function getRouteKeyName()
    {
        return 'uuid';
    }
}

// Product update route
Route::put('products/{product}', function (Request $request, Product $product) {
    $product->update($request->only('name'));

    return response()->json($product);
})->name('products.update');

// ---

route('products.update', Product::first()); // '/products/fef34a82-770a-4482-8338-10fa4cec5d50'
```

On the JavaScript side (these both work):

```js
// Using the parameter value
route('products.update', 'fef34a82-770a-4482-8338-10fa4cec5d50');

// Using an object with parameter names and values
route('products.update', { product: 'fef34a82-770a-4482-8338-10fa4cec5d50' });
```

But what if you have a JavaScript object that represents a Product model instance? Ziggy was always able to handle this _if_ the model used its `id` as its route key, but that was hard-coded just for `id`. The route model binding feature, however, passes custom bindings to the frontend, which makes this possible:

```js
const product = {
    uuid: 'fef34a82-770a-4482-8338-10fa4cec5d50',
};

// Ziggy knows to use the `uuid` key on any object passed
// in as the `product` parameter for this particular route
route('products.update', product);
```

So, circling back to your enum example in this PR—I'm unclear about what the route 'key' would be when using an enum. From your test:

```php
enum Section: string implements UrlRoutable
{
    case First = 'first';
    case Second = 'second';
    case Third = 'third';

    public function getRouteKey(): string
    {
        return $this->value;
    }

    public function getRouteKeyName(): string
    {
        return 'section';
    }
}

Route::get('sections/{section}', fn (Section $section) => '')->name('sections');
```

I'm pretty sure the `getRouteKey` and `getRouteKeyName` methods on that enum don't do anything. [Laravel never calls them.](https://github.com/laravel/framework/pull/40281/files#diff-614b34c8eaac77de1d4faf100a7133e526cfe1350ce4f02e9f60c24a88179c77R83-R87) The route key can only ever be the string backing value, so regardless of what you put in that method it would always be the `value` property of the enum case, and because of that the route key name is technically always `'value'` (but that's not a perfect comparison because it's not really a route key name).

Both of these examples should already work fine with Ziggy:

```js
route('sections', 'first'); // '/sections/first'

route('sections', { section: 'second' }); // '/sections/second'
```

What would the JavaScript code that needed a 'binding' key look like?

```js
const section = {
    [key]: 'third', // What is this key? Does this object even make sense?
};

route('sections', section);
```

Would you ever actually have a JavaScript object representing one specific enum case, instead of the actual value of that case (`'third'`)? If you did, wouldn't the key always be `'value'`?

Does this make sense? Thoughts?

bakerkretzmar:
@andrzejkupczyk no worries, thanks a lot for all of your work on this PR!

PR #507, opened 2022-01-31
still encode other characters in last parameter

rodrigopedra:
version 1.4.3 added the ability to use forward slashes on the last parameter when the user has configured a route to accept any value on it.

But other special characters should still be URL encoded.

This PR:

- Use `encodeURIComponent` to encode other characters than `/` in the last parameter
- Add a test assertion that fails without this PR's code change and passes afterwards


bakerkretzmar:
Thanks!

PR #512, opened 2022-02-16
[Vue] Make the global mixin property configurable and add doc

Tofandel:
Using 'route' as a global mixin name is a very bad idea as it conflicts with a lot of plugins out there (since it's a global mixin and just using a prop named 'route' anywhere would just result in a conflict)

It's usually recommended to prefix it with '$' for global mixins so that it wouldn't conflict with props or methods

To not make it a breaking change I just propose to make it configurable in the plugin options so that it can be changed if a conflict occurs

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,7 +354,7 @@ import App from './App';
 
 createApp(App).use(ZiggyVue, Ziggy);
 
-// Vue 2
+// Vue 2 & 3
 import Vue from 'vue'
 import { ZiggyVue } from 'ziggy';
 import { Ziggy } from './ziggy';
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
Thanks for the PR. I don't think it's worth making this configurable, the Vue plugin itself is tiny so if users want a property that isn't called `route` it's straightforward to just make their own mixin instead:

```diff
- import { ZiggyVue } from 'ziggy';
import { Ziggy } from './ziggy';

- Vue.use(ZiggyVue, Ziggy)
+ Vue.mixin({
+     methods: {
+         '$route': (name, params, absolute, config = Ziggy) => route(name, params, absolute, config)
+     }
+ })
```

To disable global registration I think it's also easier to just not use the plugin.

PR #514, opened 2022-02-16
fix: updated the route matching logic to take the wheres into account (#513)

### LINKED ISSUE #513: route().current is inconsistent with laravel
### Ziggy version

v1.4.3

### Laravel version

any

### Description

Go to the `seminars.quote` page and follow the Ziggy call

### Ziggy call and context

```js
console.log(route().params); // logs "devis"
console.log(route().current('seminars.quote', route().params)); // logs false but should be true
```

### Ziggy configuration

```json
{
  "url": "http://127.0.0.1:8000",
  "port": 8000,
  "defaults": {},
  "routes": {
    "seminars": {
      "uri": "colloques/{page?}",
      "methods": [
        "GET",
        "HEAD"
      ],
      "wheres": {
        "page": "[0-9]+"
      }
    },
    "seminars.quote": {
      "uri": "colloques/devis",
      "methods": [
        "GET",
        "HEAD"
      ]
    }
  }
}
```

### Route definition

```php
Route::prefix('colloques')->group(function () {
    Route::get('{page?}', function () {
        return "Seminars";
    })->where('page', '[0-9]+')->name('seminars');

    Route::get('devis', function () {
        return "Seminars quote";
    })->name('seminars.quote');
});
```

### Analysis of the bug

Looking at the [`matchesUrl`](https://github.com/tighten/ziggy/blob/main/src/js/Route.js#L57) method, it looks like it's not using the `wheres` property of the route, but it should validate the found route against it, otherwise it matches the parent route which has an optional parameter

A workaround is to just invert the route definition so that the one without parameter gets matched first

```php
Route::prefix('colloques')->group(function () {
    Route::get('devis', function () {
        return "Seminars quote";
    })->name('seminars.quote');
    
    Route::get('{page?}', function () {
        return "Seminars";
    })->where('page', '[0-9]+')->name('seminars');
});
```

But for more complex scenarios with overlapping params, this will not be possible, in any case the behavior differs from laravel

Tofandel:
Updated the route matching logic to take the wheres into account, completed a todo and improved performance,
I removed the _dehydrate part which became useless since all this now happens in the regex matching

Fixes #513 

- [x] Thorough tests 
- [x] ~~Do we need to support a few old browsers ? ([which don't support](https://caniuse.com/mdn-javascript_builtins_regexp_named_capture_groups) named capturing regex groups), if yes then it's possible to implement a workaround to named groups quite easily~~

### DIFF
--- a/src/js/Route.js
+++ b/src/js/Route.js
@@ -1,3 +1,5 @@
+import {parse} from "qs";
+
 /**
  * A Laravel route. This class represents one route and its configuration and metadata.
  */
@@ -52,7 +54,7 @@ export default class Route {
      * Get whether this route's template matches the given URL.
      *
      * @param {String} url - URL to check.
-     * @return {Boolean} Whether this route matches.
+     * @return {Array|false} If this route matches, returns the matched parameters.
      */
     matchesUrl(url) {
         if (!this.definition.methods.includes('GET')) return false;
@@ -60,15 +62,15 @@ export default class Route {
         // Transform the route's template into a regex that will match a hydrated URL,
         // by replacing its parameter segments with matchers for parameter values
         const pattern = this.template
-            .replace(/\/{[^}?]*\?}/g, '(\/[^/?]+)?')
-            // TODO: the above line with the leading slash is necessary to pick up completely optional *segments*,
-            // like in `/pages/{subPage?}`, so that those are handled first before the more permissive patterns
-            // below, but there's probably a way to do this in one shot
-            .replace(/{[^}?]*\?}/g, '([^/?]+)?')
-            .replace(/{[^}]+}/g, '[^/?]+')
+            .replace(/(\/?){([^}?]*)(\??)}/g, (_, slash, segment, optional) => {
+              const regex = '(?<'+segment+'>' + (this.wheres[segment] || '[^/?]+') + ')';
+              return optional ? '(' + slash + regex + ')?' : slash + regex
+            })
             .replace(/^\w+:\/\//, '');
 
-        return new RegExp(`^${pattern}$`).test(url.replace(/\/+$/, '').split('?').shift());
+        const split = url.replace(/^\w+:\/\//, '').split('?');
+        const matches = new RegExp(`^${pattern}/?$`).exec(split[0]);
+        return matches ? {params: matches.groups, query: parse(split[1])} : false;
     }
 
     /**
@@ -78,16 +80,24 @@ export default class Route {
      * @return {String}
      */
     compile(params) {
-        if (!this.parameterSegments.length) return this.template;
+        const segments = this.parameterSegments;
+        if (!segments.length) {
+            return this.template;
+        }
 
-        return this.template.replace(/{([^}?]+)\??}/g, (_, segment) => {
+        return this.template.replace(/{([^}?]+)(\??)}/g, (_, segment, optional) => {
             // If the parameter is missing but is not optional, throw an error
-            if ([null, undefined].includes(params[segment]) && this.parameterSegments.find(({ name }) => name === segment).required) {
+            if (!optional && [null, undefined].includes(params[segment])) {
                 throw new Error(`Ziggy error: '${segment}' parameter is required for route '${this.name}'.`)
             }
 
-            if (this.parameterSegments[this.parameterSegments.length - 1].name === segment && this.wheres[segment] === '.*') {
+            if (segments[segments.length - 1].name === segment && this.wheres[segment] === '.*') {
                 return encodeURIComponent(params[segment] ?? '').replaceAll('%2F', '/');
+            }
+
+            if (this.wheres[segment] &&
+                !new RegExp('^'+(optional ? '(' + this.wheres[segment] + ')?' : this.wheres[segment])+'$').test(params[segment] ?? '')) {
+                throw new Error(`Ziggy error: '${segment}' parameter does not match required format '${this.wheres[segment]}' for route '${this.name}'.`)
             }
 
             return encodeURIComponent(params[segment] ?? '');
--- a/src/js/Router.js
+++ b/src/js/Router.js
@@ -1,4 +1,4 @@
-import { parse, stringify } from 'qs';
+import { stringify } from 'qs';
 import Route from './Route';
 
 /**
@@ -6,10 +6,10 @@ import Route from './Route';
  */
 export default class Router extends String {
     /**
-     * @param {String} name - Route name.
-     * @param {(String|Number|Array|Object)} params - Route parameters.
-     * @param {Boolean} absolute - Whether to include the URL origin.
-     * @param {Object} config - Ziggy configuration.
+     * @param {String} [name] - Route name.
+     * @param {(String|Number|Array|Object)} [params] - Route parameters.
+     * @param {Boolean} [absolute] - Whether to include the URL origin.
+     * @param {Object} [config] - Ziggy configuration.
      */
     constructor(name, params, absolute = true, config) {
         super();
@@ -37,6 +37,10 @@ export default class Router extends String {
      * @return {String}
      */
     toString() {
+        if (!this._params) {
+            return window.location.toString();
+        }
+
         // Get parameters that don't correspond to any route segments to append them to the query
         const unhandled = Object.keys(this._params)
             .filter((key) => !this._route.parameterSegments.some(({ name }) => name === key))
@@ -53,6 +57,37 @@ export default class Router extends String {
     }
 
     /**
+     * Find the first route that matches the URL
+     *
+     * @param {string} [url] The url to resolve, defaults to the current url
+     * @returns {{name: string, params: Object, query: Object, route: Route}}
+     */
+    unresolve(url) {
+        if (!url) {
+            url = this._currentUrl();
+        } else if (this._config.absolute && url.startsWith('/')) {
+            const loc = this._location();
+            url += loc.host;
+        }
+        let matchedParams = {};
+        const [name, route] = Object.entries(this._config.routes).find(
+          (entry) => (matchedParams = new Route(entry[0], entry[1], this._config).matchesUrl(url))
+        ) || [undefined, undefined];
+
+        const {query, params} = matchedParams;
+        return {name, params, query, route};
+    }
+
+    _currentUrl() {
+        const loc = this._location();
+
+        return (this._config.absolute ? loc.host + loc.pathname :
+            loc.pathname.replace(this._config.url.replace(/^\w*:\/\/[^/]+/, ''), '').replace(/^\/+/, '/')
+        ) + loc.search;
+
+    }
+
+    /**
      * Get the name of the route matching the current window URL, or, given a route name
      * and parameters, check if the current window URL and parameters match that route.
      *
@@ -64,36 +99,36 @@ export default class Router extends String {
      * route().current('posts.show', { post: 1 }); // false
      * route().current('posts.show', { post: 4 }); // true
      *
-     * @param {String} name - Route name to check.
-     * @param {(String|Number|Array|Object)} params - Route parameters.
+     * @param {String} [name] - Route name to check.
+     * @param {(String|Number|Array|Object)} [params] - Route parameters.
      * @return {(Boolean|String|undefined)}
      */
     current(name, params) {
-        const url = this._config.absolute
-            ? this._location().host + this._location().pathname
-            : this._location().pathname.replace(this._config.url.replace(/^\w*:\/\/[^/]+/, ''), '').replace(/^\/+/, '/');
+        const {name: current, params: currentParams, query, route} = this.unresolve();
 
-        // Find the first route that matches the current URL
-        const [current, route] = Object.entries(this._config.routes).find(
-            ([_, route]) => new Route(name, route, this._config).matchesUrl(url)
-        ) || [undefined, undefined];
+        const matchedParams = {...currentParams, ...query};
 
-        // If a name wasn't passed, return the name of the current route
-        if (!name) return current;
+        if (!name) {
+          return current;
+        }
 
         // Test the passed name against the current route, matching some
         // basic wildcards, e.g. passing `events.*` matches `events.show`
         const match = new RegExp(`^${name.replace(/\./g, '\\.').replace(/\*/g, '.*')}$`).test(current);
 
-        if ([null, undefined].includes(params) || !match) return match;
+        if ([null, undefined].includes(params) || !match) {
+          return match;
+        }
 
         const routeObject = new Route(current, route, this._config);
+        const routeParams = JSON.parse(JSON.stringify(matchedParams)); // Remove undefined params
 
         params = this._parse(params, routeObject);
-        const routeParams = this._dehydrate(route);
 
         // If the current window URL has no route parameters, and the passed parameters are empty, return true
-        if (Object.values(params).every(p => !p) && !Object.values(routeParams).length) return true;
+        if (Object.values(params).every(p => !p) && !Object.values(routeParams).length) {
+          return true;
+        }
 
         // Check that all passed parameters match their values in the current window URL
         // Use weak equality because all values in the current window URL will be strings
@@ -126,7 +161,8 @@ export default class Router extends String {
      * @return {Object}
      */
     get params() {
-        return this._dehydrate(this._config.routes[this.current()]);
+        const {params, query} = this.unresolve();
+        return {...params, ...query};
     }
 
     /**
@@ -163,7 +199,7 @@ export default class Router extends String {
         if (Array.isArray(params)) {
             // If the parameters are an array they have to be in order, so we can transform them into
             // an object by keying them with the template segment names in the order they appear
-            params = params.reduce((result, current, i) => !!segments[i] ? ({ ...result, [segments[i].name]: current }) : ({ ...result, [current]: '' }), {});
+            params = params.reduce((result, current, i) => segments[i] ? ({ ...result, [segments[i].name]: current }) : typeof current === 'object' ? ({ ...result, ...current }) : ({ ...result, [current]: '' }), {});
         } else if (
             segments.length === 1
             && !params[segments[0].name]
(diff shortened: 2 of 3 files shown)


src/js/Router.js
```diff
@@ -88,9 +106,9 @@ export default class Router extends String {
         if ([null, undefined].includes(params) || !match) return match;
 
         const routeObject = new Route(current, route, this._config);
+        const routeParams = JSON.parse(JSON.stringify({...rParams, ...query})); // Remove undefined params
```
bakerkretzmar, line 109:
```suggestion
        const routeParams = JSON.parse(JSON.stringify(matchedParams)); // Remove undefined params
```

Also, is removing the undefined params actually necessary here?

src/js/Router.js
```diff
@@ -64,22 +78,26 @@ export default class Router extends String {
...
         // Find the first route that matches the current URL
+        let matchedParams = null;
         const [current, route] = Object.entries(this._config.routes).find(
-            ([_, route]) => new Route(name, route, this._config).matchesUrl(url)
+            ([_, route]) => (matchedParams = new Route(name, route, this._config).matchesUrl(url))
         ) || [undefined, undefined];
 
+        const {_query: query, ...rParams} = matchedParams;
+        matchedParams = {...rParams, ...query};
+        // TODO params should probably return the query params as _query, same way they are passed to compile, but would be a breaking change
```
bakerkretzmar, line 95: Would this accomplish the same thing?

```suggestion
        matchedParams = { ...matchedParams.rParams, ...matchedParams._query };
```

src/js/Route.js
```diff
@@ -52,23 +54,23 @@ export default class Route {
...
-            .replace(/\/{[^}?]*\?}/g, '(\/[^/?]+)?')
-            // TODO: the above line with the leading slash is necessary to pick up completely optional *segments*,
-            // like in `/pages/{subPage?}`, so that those are handled first before the more permissive patterns
-            // below, but there's probably a way to do this in one shot
-            .replace(/{[^}?]*\?}/g, '([^/?]+)?')
-            .replace(/{[^}]+}/g, '[^/?]+')
+            .replace(/(\/?){([^}?]*)(\??)}/g, (_, slash, segment, optional) => {
+              const regex = '(?<'+segment+'>' + (this.wheres[segment] || '[^/?]+') + ')';
+              return optional ? '(' + slash + regex + ')?' : slash + regex
+            })
```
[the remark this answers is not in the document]
Tofandel: It would just be pushing the segment name in an array and then match the matched indexes to the segment name, but it does also mean that we need to remove matching groups from the original regex if any (same way laravel does)

src/js/Router.js
```diff
@@ -11,7 +11,7 @@ export default class Router extends String {
      * @param {Boolean} absolute - Whether to include the URL origin.
      * @param {Object} config - Ziggy configuration.
      */
-    constructor(name, params, absolute = true, config) {
+    constructor(name = '', params = {}, absolute = true, config) {
```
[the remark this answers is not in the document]
Tofandel: Just for correct typing, otherwise IDE complains that the params are required

src/js/Router.js
```diff
@@ -37,6 +37,10 @@ export default class Router extends String {
      * @return {String}
      */
     toString() {
+        if (!this._params) {
+            return this._currentUrl(true)
+        }
+
```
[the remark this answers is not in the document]
Tofandel: I guess it's not

src/js/Router.js
Tofandel: Test `same(route().current('pages.optionalExtension', ''), true);` fails without it

src/js/Router.js
```diff
@@ -107,9 +125,10 @@ export default class Router extends String {
      * @return {Object}
      */
     _location() {
-        const { host = '', pathname = '', search = '' } = typeof window !== 'undefined' ? window.location : {};
+        const { protocol = '', host = '', pathname = '', search = '' } = typeof window !== 'undefined' ? window.location : {};
```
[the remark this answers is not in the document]
Tofandel: Removed

.gitignore
```diff
@@ -4,3 +4,4 @@ vendor/
 *.log
 composer.lock
 phpunit.xml
+.idea
```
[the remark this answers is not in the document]
bakerkretzmar: IDE files like this should go in your global gitignore, Ziggy's gitignore is for files that are specific to Ziggy that will always be present but should be ignored.

src/js/Router.js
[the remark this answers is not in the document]
bakerkretzmar: Okay thanks, let's revert this then please. Not sure why your IDE says they're required, but they aren't.

src/js/Router.js
bakerkretzmar: Ah right okay, thanks!

src/js/Route.js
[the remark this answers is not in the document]
bakerkretzmar: Okay great. I don't care about IE 11, and Safari 11 usage is tiny so let's keep this!

src/js/Router.js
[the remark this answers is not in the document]
Tofandel: It's hinting they are because the doc comments doesn't say it's optional and the function signature as well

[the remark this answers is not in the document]
Tofandel: I fixed the JS Doc instead to mark them optional https://jsdoc.app/tags-param.html#optional-parameters-and-default-values

Tofandel:
I made the logic to search the current route accessible under the new function `resolve` so that url can be converted into route names with params and query strings from a core api which wasn't possible before

.github/workflows/tests.yml
```diff
@@ -18,6 +18,8 @@ jobs:
             laravel: 'dev-master as 9.0'
           - php: '7.4'
             laravel: 'dev-master as 9.0'
+          - php: '8.0'
```
Tofandel, line 21: Laravel 9.1 requires php 8.1 hence tests where failing

bakerkretzmar:
@Tofandel thanks a lot for your hard work on this PR. I'm having a bit of trouble keeping track of all the changes at this point, I'd really like to consider all of the other ideas you have here but I think the individual PRs are going to have to be smaller so they can each be focused on one specific feature or fix.

Could you please update this PR to include only the parts specific to making `wheres` work, mainly the new regex with named groups and the removal of the `_dehydrate` method, and submit the `resolve`/`unresolve` parts as a separate PR?

Also, please just push additional commits to this branch instead of force pushing so that I can see the history clearly and see any test failures along the way. Thanks!

.github/workflows/tests.yml
bakerkretzmar: Laravel 9 actually only requires PHP 8.0, what test failures did you see?

Tofandel: Dev master is now 9.1 which requires 8.1, just trigger a travis build and you'll see

Tofandel: https://github.com/laravel/framework/commit/c4271001aa15a34af41b8205674ce62cd6c7aaaf

Tofandel: Ideally we could now just put 9.0 instead of dev master but it was just a quick fix to get the tests working again

Tofandel:
Well the unresolve is 100% intertwinned with this PR, it makes no sense to separate it as this fix requires it. The resolve was added to stay consistent and to provide the opposite function of unresolve (same as route() but without the query mess) so I just encapsulated the toString code into the unresolve

.github/workflows/tests.yml
bakerkretzmar: Ahh gotcha, that's right `dev-master` is technically for Laravel 10.x. I'll fix those tests in the base branch. Thanks!

Tofandel:
I'll add some comments on the change to help you navigate it

src/js/Route.js
```diff
@@ -78,18 +80,26 @@ export default class Route {
      * @return {String}
      */
     compile(params) {
-        if (!this.parameterSegments.length) return this.template;
+        const segments = this.parameterSegments;
```
Tofandel, line 83: This whole change is just a performance improvement, to avoid calling the getter with an expensive regex 3 times

src/js/Route.js
```diff
@@ -78,18 +80,26 @@ export default class Route {
...
+            if (!optional && [null, undefined].includes(params[segment])) {
                 throw new Error(`Ziggy error: '${segment}' parameter is required for route '${this.name}'.`)
             }
 
-            if (this.parameterSegments[this.parameterSegments.length - 1].name === segment && this.wheres[segment] === '.*') {
+            if (segments[segments.length - 1].name === segment && this.wheres[segment] === '.*') {
                 return encodeURIComponent(params[segment] ?? '').replaceAll('%2F', '/');
             }
 
+            if (this.wheres[segment] &&
```
Tofandel, line 98: This change is just to throw an error if we are compiling the route with params that don't match the where format

src/js/Router.js
```diff
@@ -1,15 +1,15 @@
...
 /**
  * A collection of Laravel routes. This class constitutes Ziggy's main API.
  */
 export default class Router extends String {
     /**
-     * @param {String} name - Route name.
-     * @param {(String|Number|Array|Object)} params - Route parameters.
-     * @param {Boolean} absolute - Whether to include the URL origin.
-     * @param {Object} config - Ziggy configuration.
+     * @param {String} [name] - Route name.
```
Tofandel, line 9: Just a jsdoc change for IDE to understand the params are all optional

src/js/Router.js
```diff
@@ -18,12 +18,8 @@ export default class Router extends String {
         this._config = { ...this._config, absolute };
 
         if (name) {
-            if (!this._config.routes[name]) {
-                throw new Error(`Ziggy error: route '${name}' is not in the route list.`);
-            }
-
-            this._route = new Route(name, this._config.routes[name], this._config);
-            this._params = this._parse(params);
+            this._name = name;
```
Tofandel, line 21: I'm trying to get the public methods to not depend on `this` so only the passed params are stored and this error now happens in resolve()

src/js/Router.js
```diff
@@ -37,13 +33,35 @@ export default class Router extends String {
      * @return {String}
      */
     toString() {
+        return this.resolve({name: this._name, params: this._params});
```
Tofandel, line 36: The toString logic was moved to resolve

src/js/Router.js
```diff
@@ -37,13 +33,35 @@ export default class Router extends String {
...
+
+    /**
+     * Con
+     * @param {string} name
+     * @param {Object|string} [params]
+     * @param {Object} [query]
+     * @returns {string}
+     */
+    resolve({name, params, query}) {
+        if (!name) {
```
Tofandel, line 48: As explained this is the toString logic abstracted to not depend on the instance route and params, so that any route can be passed to it

src/js/Router.js
```diff
@@ -52,6 +71,37 @@ export default class Router extends String {
...
+        const [name, route] = Object.entries(this._config.routes).find(
+          (entry) => (matchedParams = new Route(entry[0], entry[1], this._config).matchesUrl(url))
+        ) || [undefined, undefined];
+
+        const {query, params} = matchedParams;
+        return {name, params, query, route};
+    }
+
+    _currentUrl() {
+        const loc = this._location();
```
Tofandel, line 97: Just an abstraction taken from `current()`

tests/js/route.test.js
```diff
@@ -155,11 +210,18 @@ beforeAll(() => {
 
 beforeEach(() => {
     window.location = { ...defaultWindow.location };
-    global.window.location = { ...defaultWindow.location };
+    window.location.toString = () => window.location.protocol + '//' + window.location.host + window.location.pathname + window.location.search + window.location.hash;
```
Tofandel, line 213: Just making the window mock more accurate as it was missing the toString method

tests/js/route.test.js
```diff
@@ -502,8 +564,9 @@ describe('route()', () => {
 
     test("can append 'extra' string/number parameter to query", () => {
         // 'posts.index' has no parameters
-         same(route('posts.index', 'extra'), 'https://ziggy.dev/posts?extra=');
-         same(route('posts.index', 1), 'https://ziggy.dev/posts?1=');
+        same(route('posts.index', 'extra'), 'https://ziggy.dev/posts?extra=');
+        same(route('posts.index', {extra: 2}), 'https://ziggy.dev/posts?extra=2');
```
Tofandel, line 568: I added this test because prior to this PR, this would result in 'https://ziggy.dev/posts?[Object object]=' which is a bug fixed on line https://github.com/tighten/ziggy/pull/514/files#diff-372121aa5f6bd350905f140546c9ea2736d4b4f2a6cf987115c77eba82f936ddR217

tests/js/route.test.js
```diff
@@ -533,6 +596,19 @@ describe('route()', () => {
         same(route('pages.optionalExtension', { extension: '.pdf' }), 'https://ziggy.dev/download/file.pdf');
     });
 
+    test('can generate a URL for a route with optional parameters inside individual segments respecting where requirements', () => {
```
Tofandel, line 599: All the tests to make sure the where fix is implemented correctly

tests/js/route.test.js
```diff
@@ -945,4 +1079,9 @@ describe('current()', () => {
 
         same(route(undefined, undefined, undefined, config).current(), 'events.venues.show');
     });
+
+
+    test('can resolve the same route(name, params) resolves to', () => {
```
Tofandel, line 1084: Test for resolve()

tests/js/route.test.js
```diff
@@ -919,6 +1043,15 @@ describe('current()', () => {
         same(route().current('events.venues.*'), false);
     });
 
+
+    test('can unresolve arbitrary urls to names and params', () => {
```
Tofandel, line 1047: Test for unresolve()

src/js/Router.js
```diff
@@ -37,6 +37,10 @@ export default class Router extends String {
      * @return {String}
      */
     toString() {
+        if (!this._params) {
```
Tofandel, line 40: This is to avoid a failure if calling toString when this._params is not defined, not really publicly exposed but good to have

src/js/Router.js
```diff
@@ -52,6 +56,37 @@ export default class Router extends String {
         });
     }
 
+    /**
+     * Find the first route that matches the URL
+     *
+     * @param {string} [url] The url to resolve, defaults to the current url
+     * @returns {{name: string, params: Object, query: Object, route: Route}}
+     */
+    unresolve(url) {
```
Tofandel, line 65: This is also an abstraction of current() so that you can retrieve any route, params and query just from its url

src/js/Router.js
```diff
@@ -64,36 +99,36 @@ export default class Router extends String {
...
+     * @param {(String|Number|Array|Object)} [params] - Route parameters.
      * @return {(Boolean|String|undefined)}
      */
     current(name, params) {
-        const url = this._config.absolute
-            ? this._location().host + this._location().pathname
-            : this._location().pathname.replace(this._config.url.replace(/^\w*:\/\/[^/]+/, ''), '').replace(/^\/+/, '/');
+        const {name: current, params: currentParams, query, route} = this.unresolve();
 
-        // Find the first route that matches the current URL
```
Tofandel, line 76: This is what was put into resolve()

src/js/Router.js
```diff
@@ -64,36 +99,36 @@ export default class Router extends String {
...
      * route().current('posts.show', { post: 4 }); // true
      *
-     * @param {String} name - Route name to check.
-     * @param {(String|Number|Array|Object)} params - Route parameters.
+     * @param {String} [name] - Route name to check.
+     * @param {(String|Number|Array|Object)} [params] - Route parameters.
      * @return {(Boolean|String|undefined)}
      */
     current(name, params) {
-        const url = this._config.absolute
```
Tofandel, line 72: What was put into _currentUrl()

src/js/Router.js
```diff
@@ -163,7 +199,7 @@ export default class Router extends String {
         if (Array.isArray(params)) {
             // If the parameters are an array they have to be in order, so we can transform them into
             // an object by keying them with the template segment names in the order they appear
-            params = params.reduce((result, current, i) => !!segments[i] ? ({ ...result, [segments[i].name]: current }) : ({ ...result, [current]: '' }), {});
+            params = params.reduce((result, current, i) => segments[i] ? ({ ...result, [segments[i].name]: current }) : typeof current === 'object' ? ({ ...result, ...current }) : ({ ...result, [current]: '' }), {});
```
Tofandel, line 202: Fix a bug where route('name', {unexistingParam: value}) would return `name?[Object object]=`

Tofandel:
I rebased without the resolve method which I'll do under another PR, added relevant comments in this one

.github/workflows/tests.yml
```diff
@@ -12,12 +12,12 @@ jobs:
     strategy:
       matrix:
         php: ['7.3', '7.4', '8.0', '8.1']
-        laravel: [5.5.*, 5.6.*, 5.7.*, 5.8.*, 6.*, 7.*, 8.*, 'dev-master as 9.0']
+        laravel: [5.5.*, 5.6.*, 5.7.*, 5.8.*, 6.*, 7.*, 8.*, 9.*]
```
Tofandel, line 15: Fix tests with laravel dev-master, feel free to cherry pick the second commit has it has only this change or to fix it on main so I can rebase

src/js/Router.js
bakerkretzmar: Nice, didn't know about this. Thanks.

bakerkretzmar (commented):
@Tofandel this looks great and it's almost ready, thanks so much again! I left some comments with a few small formatting things and changes that aren't relevant to `wheres` that I'd like to remove, and then I think this is good to go! 🙌🏻 

**Please do not force push**, just add additional commits. Force pushes make this a lot harder to review because I can't see what actually changed so I have to look through the whole diff again.

tests/js/route.test.js
```diff
@@ -155,11 +210,18 @@ beforeAll(() => {
 
 beforeEach(() => {
     window.location = { ...defaultWindow.location };
-    global.window.location = { ...defaultWindow.location };
+    window.location.toString = () => window.location.protocol + '//' + window.location.host + window.location.pathname + window.location.search + window.location.hash;
+    global.window.location = window.location;
```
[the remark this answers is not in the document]
Tofandel: But why would it be a new object? A reference is better, they should point to the same thing no so they both get the same updates and values no?

tests/js/route.test.js
```diff
@@ -667,6 +743,48 @@ describe('current()', () => {
...
+        deepEqual(route().params, {digit: '12', word: 'complex', required: 'required', optional: 'optional', extension: undefined})
+
+        global.window.location.pathname = '/where/123-abc/required/file.html';
+        deepEqual(route().params, {digit: '123', word: 'abc', required: 'required', optional: undefined, extension: '.html'})
+
+        global.window.location.pathname = '/where/complex-12/different_but_required/optional/file';
+        deepEqual(route().params, {digit: '12', required: 'different_but_required', optional: 'optional', extension: undefined})
+
+        global.window.location.search = '?ab=cd&ef=1&dd';
+        deepEqual(route().params, {digit: '12', required: 'different_but_required', optional: 'optional', extension: undefined, ab: 'cd', ef: '1', 'dd': ''})
```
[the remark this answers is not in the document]
Tofandel: `route('name', [{ extra: 2 }])` and `route('name', { extra: 2 })` should both give the same result (before PR and after PR) except before they both would give `?[Object object]=`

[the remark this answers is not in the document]
bakerkretzmar: `route('name', { extra: 2 })` worked fine before, it just added a `?extra=2` query. The bug you noticed is specifically if the whole parameters argument is an array.

tests/js/route.test.js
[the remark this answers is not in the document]
bakerkretzmar: Hmm, yeah actually I think you're right. Not sure why I had it the other way before. Thanks.

tests/js/route.test.js
Tofandel: route('name', [{ extra: 2 }]) and route('name', { extra: 2 }) should both give the same result (before PR and after PR) except before they both would give ?[Object object]=

tests/js/route.test.js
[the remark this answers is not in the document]
Tofandel: I see then let me change that test

Tofandel:
All done

bakerkretzmar:
Awesome, thanks so much @Tofandel!!

PR #517, opened 2022-02-16
refactor: add eslint

Tofandel:
Was kindy weird to be editing without a set of formatting rule, it also caught a few errors while developing so this should be a good addition

Rebased on #521 see https://github.com/Tofandel/ziggy/pull/2/files for real diffs


bakerkretzmar:
Not interested in adding eslint right now. Appreciate the work on this PR though 🙏🏻 thanks!

PR #518, opened 2022-02-22
feat: Add Vue3 Provide/Inject Plugin

c-fitzmaurice:
This PR adds a new plugin for Vue3 using Ziggy via the Provide/Inject Composition API. Using the [advanced Vue set up](https://github.com/tighten/ziggy#vue), I ran into my problem when trying to use `route()` in `setup()`. I saw a possible solution in https://github.com/tighten/ziggy/issues/454#issuecomment-906418423 but it didn't feel like a super maintainable solution, also `mixins` are [not recommended](https://vuejs.org/api/application.html#app-mixin) in Vue3 anymore.

> Note: I'm not familiar with microbundles, so if anything needs to be edited in package.json, please let me know.

For this to work, you have to update your `mix` alias like below:

```javascript
mix.alias({
    ziggy: path.resolve('vendor/tightenco/ziggy/dist/vue3'), 
});
```

and the import of `ZiggyVue3`:

```javascript
import { createApp } from 'vue';
import { ZiggyVue3 } from 'ziggy';
import App from './App';

createApp(App).use(ZiggyVue3, Ziggy);
```

The `route()` helper will still be registered globally and you gain the below functionality:

```javascript
<script>
import { inject } from 'vue';

export default {
    setup() {
        const route = inject('route');

        const navigation = [{
            name: 'Profile',
            href: route('profile'),
            current: route().current('profile'),
        }, {
            name: 'Settings',
            href: route('settings'),
            current: route().current('settings'),
        }, {
         ...
        }];

        return {
            navigation,
        };
    },
};
</script>
```

### DIFF
--- a/src/js/vue.js
+++ b/src/js/vue.js
@@ -1,9 +1,11 @@
 import route from './index.js';
 
 export const ZiggyVue = {
-    install: (v, options) => v.mixin({
-        methods: {
-            route: (name, params, absolute, config = options) => route(name, params, absolute, config),
-        },
-    }),
+    install: (v, options) => {
+        const version = parseInt(v.version.split('.')[0]);
+        const _route = (name, params, absolute, config = options) => route(name, params, absolute, config);
+
+        v.mixin && version === 2 && (v.mixin({ methods: { route: _route } }));
+        v.provide && version > 2 && (v.provide('route', _route));
+    },
 };
--- a/src/js/vue.js
+++ b/src/js/vue.js
@@ -2,10 +2,17 @@ import route from './index.js';
 
 export const ZiggyVue = {
     install: (v, options) => {
-        const version = parseInt(v.version.split('.')[0]);
-        const _route = (name, params, absolute, config = options) => route(name, params, absolute, config);
+        const r = (name, params, absolute, config = options) => route(name, params, absolute, config);
 
-        v.mixin && version === 2 && (v.mixin({ methods: { route: _route } }));
-        v.provide && version > 2 && (v.provide('route', _route));
+        v.mixin({
+            methods: {
+                route: r,
+            },
+        });
+
+        if (parseInt(v.version > 2)) {
+            v.provide('route', r);
+            v.config.globalProperties.$route = r;
+        }
     },
 };
--- a/src/js/vue.js
+++ b/src/js/vue.js
@@ -10,7 +10,7 @@ export const ZiggyVue = {
             },
         });
 
-        if (parseInt(v.version > 2)) {
+        if (parseInt(v.version) > 2) {
             v.provide('route', r);
         }
     },
--- a/src/js/vue.js
+++ b/src/js/vue.js
@@ -12,7 +12,6 @@ export const ZiggyVue = {
 
         if (parseInt(v.version > 2)) {
             v.provide('route', r);
-            v.config.globalProperties.$route = r;
         }
     },
 };


src/js/vue3.js
```diff
@@ -0,0 +1,10 @@
+import route from './index.js';
+
+export const ZiggyVue3 = {
+    install: (app, options) => {
+        const _route = (name, params, absolute, config = options) => route(name, params, absolute, options);
+
+        app.config.globalProperties.route = _route;
```
bakerkretzmar, line 7: What exactly does this line do? Adding it to `globalProperties` specifically.

bakerkretzmar:
Thanks for the PR! Making `inject` work out of the box like this would be awesome. Would it be possible to accomplish this using our existing Vue plugin? Could be just tweak it a bit and add `app.provide('route', route)`?

c-fitzmaurice:
@bakerkretzmar I can def give it a shot to add it to the existing Vue plugin, but on the [Vue 2 docs](https://v2.vuejs.org/v2/api/?redirect=true#provide-inject) the provide/inject API is only supported in `v2.2.0+`. If you're not worried about that requirement, I'll circle back later today!

src/js/vue3.js
c-fitzmaurice: Without this line, a user would have to have to inject `route()` into every component, even if they don't need it in the `setup()` function. This allows `route()` to be used in templates globally.

The below would have to be in every Vue component that wants to use `route()`.

```javascript
<script>
import { inject } from 'vue';
export default {
    setup() {
        const route = inject('route');
        return { route };
    },
};
</script>
```

c-fitzmaurice: Serving as the `mixin` to make the function globally available to Vue.

c-fitzmaurice:
@bakerkretzmar After spinning up a Vue2 Inertiajs demo, I don't think it will be possible with one plugin. I believe that the Vue2 provide/inject is only from parent to child component and not possible globally. I could be wrong if you have better Vue2 chops than I do and want to give it a shot, sorry.

Tofandel:
I was about to make a similar PR so I'll add my 2 grains of salt, it's definitely possible to support it in a single plugin

The prefered method is to export a "useRoute" method which will return the route method created when installing the plugin

Tofandel:
This is pretty much how:

```js
import route from './index.js';

let installedRoute = null;

export const ZiggyVue = {
  install: (v, options, property = 'route') => {
    installedRoute = (name, params, absolute, config = options) => route(name, params, absolute, config);
    v.mixin({
      methods: {
        [property]: installedRoute,
      },
    });
  },
};


export const useRoute = () => installedRoute;
```

src/js/vue3.js
bakerkretzmar: Do you know if the `globalProperties` thing works _exactly_ the same as mixins? Can you still call `this.route()` inside components using the Options API?

bakerkretzmar:
I think we can probably get this working without a separate Vue 3 plugin, Vue has a `version` property we can check. Something like `import { version } from 'vue'` and then if we're on Vue 3, we call `app.provide()`.

I don't want to split it out into two different plugins if possible.

Tofandel:
@bakerkretzmar Well, I have a version not using inject provide that does that, I'd like to use inject provide as well, the problem is with the `ìmport { inject } from 'vue';` because we can't have dynamic imports and the ìnject` doesn't exist in vue 2 it woud likely throw an error

bakerkretzmar:
Not `import { inject } from 'vue'`, `import { version } from 'vue'`. `version` exists in Vue 2 and 3.

Tofandel:
I don't think you understood the issue, the problem is with `microbundle --external none` then if we `import { version } from vue` then it won't matter the version because it will bundle vue and use the vue installed in ziggy, which is a problem. This is what I'm saying, anything importing something from vue is a problem with this configuration

Tofandel:
I don't think you understood the issue, the problem is with `microbundle --external none` then if we `import { version } from vue` then it won't matter the version because it will bundle vue and use the vue installed in ziggy, which is a problem. This is what I'm saying, anything importing something from vue is a problem with this configuration

src/js/vue3.js
Tofandel: @bakerkretzmar What do you mean by "inside components" do you mean in the template or in the script? Because if it's in the script the composition API doesn't have a `this` that's why mixins don't work in the composition API and we need a `useRoute` method

c-fitzmaurice:
Okay, so I have read up on this a little more and just pushed an update to this PR.

@Tofandel I tested your solution and it works, but it may present some uneeded complexity with the options of changing the `route` name and the ability to turn off the global mixin in favor of an import with `useRoute()`. A user can just not install the plugin into the Vue instance and import route diretly with `import route from '../../../../vendor/tightenco/ziggy/dist';` if they want to.

In the new implementation I just pushed I also decided to remove the below line since the new version leaved the Vue 2 implementation unaffected.

`app.config.globalProperties.route = _route;`

The new version allows the Vue 2 implementation to work without any breaking changes and provides Vue 3 with Provide/Inject, also without chaning the installation steps. Mixins are available in Vue 3 but there are many places in the docs saying they are no longer recomended. There are many blog posts out there listing the reasons why. I tested all the below implementations and can confirm they work. So this version provides a clear break in the plugin to align with Vue best practices, Vue 2 - Mixin is okay, Vue 4 Provide/Inject.

> Note: There are several ways to check if `provide` is undefined or parsing out the Vue version. If you prefer another syntax just let me know. Also, `version > 2` was an attempt to potentially future proof it a little assuming the composition API is here to stay.

## Vue 2 - Unaffected

```js
<template>
    <a :href="route('home')">route</a>
    <a :href="url">url</a>
</template>

<script>
export default {
    data() {
        return {
            url: this.route('home'),
        }
    },
};
</script>
```

## Vue 3 - Options API
```js
<template>
    <a :href="route('home')">route</a>
    <a :href="url">url</a>
</template>

<script>
export default {
    inject: ['route'],
    data() {
        return {
          url: this.route('home'),
        }
    }
}
</script>
```

## Vue 3 - Composition API

```js
<template>
    <a :href="route('home')">route</a>
    <a :href="url">url</a>
</template>

<script setup>
import { inject } from 'vue';
const route = inject('route');

const url = route('home');
</script>
```

src/js/vue3.js
bakerkretzmar: Yeah I just meant using `this.route()` inside the script part of a component using the Options API.

src/js/vue.js
```diff
@@ -1,9 +1,11 @@
...
-    install: (v, options) => v.mixin({
-        methods: {
-            route: (name, params, absolute, config = options) => route(name, params, absolute, config),
-        },
-    }),
+    install: (v, options) => {
+        const version = parseInt(v.version.split('.')[0]);
+        const _route = (name, params, absolute, config = options) => route(name, params, absolute, config);
+
+        v.mixin && version === 2 && (v.mixin({ methods: { route: _route } }));
```
bakerkretzmar, line 8: I'm on board with this in principle but it breaks backwards compatibility for all existing Vue 3 apps using Ziggy. We have to keep the mixin for now.

src/js/vue.js
```diff
@@ -1,9 +1,11 @@
...
+        v.provide && version > 2 && (v.provide('route', _route));
```
bakerkretzmar, line 9: Why the `v.provide` check? If we're on Vue 3 `provide` is always going to be there, right?

bakerkretzmar (commented):
@c-fitzmaurice thanks a lot for digging into this further. I had another look at it last week too and landed on something very similar but forgot to come back and comment 🤦🏻 

Vue 2: https://codesandbox.io/s/ziggy-vue-2-n3x2e0
Vue 3: https://codesandbox.io/s/ziggy-vue-3-xnz66z

I left a couple comments but after they're resolved I think this is ready to go. I'm on board with removing the mixin eventually but that's a significant breaking change so it'll have to wait until v2.

src/js/vue.js
c-fitzmaurice: You're right, that can be removed, probably overkill.

src/js/vue.js
c-fitzmaurice: I'd have to check again, but if a user needs `route()` in their `<script setup>` and uses `inject()` I'm pretty sure it will conflict with the `mixin route()` loaded automatically in the `<template>`

bakerkretzmar: Interesting, "conflict" how? Would it error or would one just override the other, which shouldn't matter? I'll try it in the sandbox again and see what happens.

c-fitzmaurice: Ugh, sorry I don't remember but I pulled it last night cause it was breaking...but I may have screwed something up. If it works, we should keep it then! Thanks for helping.

If it works, I will update the PR again today adding back the mixin.

bakerkretzmar: Looks to me like it works fine?

https://stackblitz.com/edit/vitejs-vite-pdjx4q?file=src%2FSetup.vue&terminal=dev

c-fitzmaurice: Yupp, looks great. I will make the syntax tweaks and put back the mixin and push them up shortly.

c-fitzmaurice:
@bakerkretzmar Updated the PR with the code from the below as discussed.

https://stackblitz.com/edit/vitejs-vite-zuznmu

bakerkretzmar:
@c-fitzmaurice thanks a lot!

Tofandel:
The check should not have been on the version but just whether or not vue is available, as the composition api is available in vue 2 with an official package

Tofandel:
Interesting find, then I guess users of vue2 composition API will have to provide it manually in their root component

psr-cookies:
>  So this version provides a clear break in the plugin to align with Vue best practices, Vue 2 - Mixin is okay, Vue 3 - Provide/Inject.

The ideal would be to use composables with the Vue Composition API

[**What is a "Composable"?**](https://vuejs.org/guide/reusability/composables.html)

> In the context of Vue applications, a "composable" is a function that leverages Vue's Composition API to encapsulate and reuse stateful logic.

Provide / Inject serve to pass information from a parent component to another child component, without going through the parent component.

With the "Composition API" option, using Provide/Inject is not best practice in Vue 3.

PR #521, opened 2022-03-01
feat: Add resolve() method

Tofandel:
Follow up of #514 to make API consistent


bakerkretzmar:
Is there a particular use case you have for this? The code looks fine but I'm not sure it's necessary, the new `resolve()` method basically just does exactly what `toString()` does now, right?

Tofandel:
Just to make the api consistent, I had some problems with the fact params currently doesn't give query params separately, so I fixed that in unresolve which returns an object of the form that resolve takes it in, not to have to transform it

You'll notice this is also the same api as vue router :to="{name, params, query}" and that's because I made a plugin to make inertia.js +ziggy a drop In vue router replacement so that it works with the whole vue ecosystem and I use this method instead

PR #522, opened 2022-03-03
feat: Add watcher option to generate command and allow for options to be specified multiple times

Tofandel:
This is a very nice backward compatible improvement to the ziggy:generate command

It allows the command to wait until Enter is pressed meanwhile watching for file change to regenerate them, this is very usefull in conjonction with the watch option of webpack or laravel mix as it allows for a development process not requiring manually running commands

Here is an example of usage
`php artisan ziggy:generate ./resources/js/Routes/admin.js  --group=admin ./resources/js/Routes/public.js --watch`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,9 @@ You can run PHP tests with `vendor/bin/phpunit` and JavaScript tests with `npm t
 
 If you need any help with this please don't hesitate to ask.
 
+> :warning: If your filesystem uses CRLF instead of LF you may run into some issues when running the tests
+> You can run `find . -type f -not -path "./vendor" -not -path "./node_modules" -exec grep -qIP '\r$' {} ';' -exec perl -pi -e 's/\r\n/\n/g' {} '+'` to restore files back to LF
+
 ## Requirements
 
 - **[PSR-2 Coding Standard](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md)** - The easiest way to apply the conventions is to install [PHP Code Sniffer](https://github.com/squizlabs/PHP_CodeSniffer).
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,11 @@
             "email": "jacob@tighten.co"
         }
     ],
+    "suggest": {
+        "ext-inotify": "To use the watch option of the generate command"
+    },
     "require": {
+        "ext-json": "*",
         "laravel/framework": ">=5.4@dev"
     },
     "require-dev": {
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,6 +43,7 @@ jobs:
       - uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
+          extensions: inotify
       - uses: actions/cache@v2
         with:
           path: ~/.composer/cache/files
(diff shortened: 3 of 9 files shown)


Tofandel:
I'm battling in my dev environnement with the tests

The editorconfig says 

end_of_line = lf


But all the files actually have CRLF endings, and in the BladeRouteGeneratorTest some tests use '\n\n' instead of `\r\n\r\n` this problem is driving me nuts

composer.json
```diff
@@ -18,7 +18,11 @@
             "email": "jacob@tighten.co"
         }
     ],
+    "suggest": {
+        "ext-inotify": "To use the watch option of the generate command"
+    },
     "require": {
+        "ext-json": "*",
```
Tofandel, line 25: This was missing and is required to use json_encode / json_decode in php, even though this extension is usually installed by default in php

CONTRIBUTING.md
```diff
@@ -33,6 +33,9 @@ You can run PHP tests with `vendor/bin/phpunit` and JavaScript tests with `npm t
 
 If you need any help with this please don't hesitate to ask.
 
+> :warning: If your filesystem uses CRLF instead of LF you may run into some issues when running the tests
```
Tofandel, line 36: I added this because I've been struggling for a good hour with this while running tests locally

bakerkretzmar:
@Tofandel I appreciate the work you put into this but I'm going to pass for now. This feature has been suggested before but it's simpler to just run `php artisan ziggy:generate` when your PHP route files change, or set up a Mix hook as you demonstrated here.

Tofandel:
Hmm I strongly disagree the mix hook doesn't have any watchers, without this command modification it has no point at all, I'm already using this and it works really perfectly

It's really not simpler to run a command when you're developing with mix watch

bakerkretzmar:
Happy to explain further. I don't want to reach into Laravel to modify internal properties, reload configurations, reboot the app, etc. It's not a great idea to do stuff like that in general, but it's particularly risky in a package like Ziggy because app developers may not know that that's happening and it could interact with their app and with other packages in unexpected ways. That would have to be removed from this PR. Even without those parts though, it adds a lot of complexity that we'll have to maintain, and I don't think it's worth it for this feature considering that there are other ways to accomplish the same thing.

It's an interesting idea, and I really do appreciate the time you put in to contribute!

Here's an example from an earlier discussion of how to add a watcher to Mix to recompile when a route file changes: https://github.com/tighten/ziggy/issues/321#issuecomment-699159823.

It could probably be even simpler though:

```js
const { exec } = require('child_process');
const { resolve } = require('path');
const glob = require('glob');

mix.webpackConfig({
    plugins: [
        {
            apply: (compiler) => {
                compiler.hooks.compilation.tap('WatchRoutesPlugin', (compilation) => {
                    glob.sync('./routes/*.php').map(file => compilation.fileDependencies.add(resolve(file)));
                });
                compiler.hooks.afterDone.tap('ZiggyGeneratePlugin', () => {
                    exec('php artisan ziggy:generate');
                });
            }
        },
    ],
});
```

Tofandel:
Well, then maybe ask for changes instead of a categoric no, there is other ways to reload the config, like create a new app that doesn't involve anything internal same way it's done in the test suite

I already tried to implement a watcher in js, but it just doesn't work because it only executes after a change in js files and so it goes like this => adding a feature in js with a route name all happy you go to check it on the front, dammit forgot to name my route => go to name the route, reload the page, dammit it doesn't recompile, you have to change the js for it to recompile again, in the end, in the end the experience is terrible

Tofandel:
Okay sorry I misunderstood this code

I could modify this PR then, and actually provide a webpack plugin that gives the feature like proposed as I still really would like the multiple arguments part

PR #529, opened 2022-03-13
feat: allow options multiple times in generate command

Tofandel:
Split from #522

This PR allows for multiple options to be given to generate multiple files at once


bakerkretzmar:
Appreciate your work on this but I'm going to pass on adding it to Ziggy for now. If you need to generate multiple files, run the command multiple times, if you need to do that every time, put the commands in a Composer script.

PR #547, opened 2022-04-05
feat: expose route method

jaulz:
This PR restructures `Ziggy` class so it can expose a `route` method (name is up to discussion) that returns a route in the same format as it is exposed in the generated file. In my use case I use it to expose the current route separately:
```php
        $ziggy = new \Tightenco\Ziggy\Ziggy();

        $ziggy
          ->route($request->route())
          ->toArray();
```

### DIFF
--- a/src/Ziggy.php
+++ b/src/Ziggy.php
@@ -94,6 +94,25 @@ class Ziggy implements JsonSerializable
     }
 
     /**
+     * Prepare route for output.
+     */
+    public function route($route)
+    {
+        $bindings = $this->resolveBindings($route);
+
+        return collect($route)->only(['uri', 'methods', 'wheres'])
+            ->put('domain', $route->domain())
+            ->put('bindings', $bindings ?? [])
+            ->when($middleware = config('ziggy.middleware'), function ($collection) use ($middleware, $route) {
+                if (is_array($middleware)) {
+                    return $collection->put('middleware', collect($route->middleware())->intersect($middleware)->values()->all());
+                }
+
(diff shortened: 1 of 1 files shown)


jaulz:
To reduce the app bundle size I use a Babel macro that replaces `route` calls with the actual route definitions (e.g. `route("home")` will be compiled into `{ "uri": "...", "methods": [...], "bindings": [] }` based on the generated JSON file by Ziggy). The only route that I would like to expose is the current route so I can easily manipulate the current URL via Ziggy. Consequently I need it in the format that is expected by the JavaScript class and thus I would simply use a method of this library instead of using my own custom function (that I might need to adapt if the library changes). 

It is a special scenario but I think the changes of this PR would not add any complexity.

jaulz:
@bakerkretzmar any chance to have another look? 😊  thanks in advance!

bakerkretzmar:
@jaulz will revisit this again this week 👍🏻 this would accomplish more or less the same thing, right?

```php
use Tightenco\Ziggy\Ziggy;

(new Ziggy)->toArray()['routes'][$request->route()];
```

jaulz:
@bakerkretzmar sorry to disturb but would you have any chance to have another look at this? 😊

jaulz:
@bakerkretzmar sorry but I will close the PR for the time being but thanks anyway for the great project! In the meantime I moved on and released a framework which is quite similar to Inertia but comes with a lot of new features including injected route definitions via Vite. In terms of route matching it's largely similar to your's, so thanks for that! You can find more details here: https://github.com/navigarejs/framework

PR #555, opened 2022-04-26
Add support for using only and except in group

YeeJiaWei:
### Discussed in https://github.com/tighten/ziggy/discussions/520

### DIFF
--- a/src/Ziggy.php
+++ b/src/Ziggy.php
@@ -19,13 +19,13 @@ class Ziggy implements JsonSerializable
     protected $group;
     protected $routes;
 
-    public function __construct($group = null, string $url = null)
+    public function __construct($group = NULL, string $url = NULL)
     {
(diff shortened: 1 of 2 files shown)


hulkur:
yes, please :) 

just came to look if it is already supported or needs a PR

YeeJiaWei:
yes, I opened an PR

bakerkretzmar:
Thanks! Couple things I noticed that I'll see if I can fix quickly:

- I believe these nested `only` and `except` configs will be ignored if a user passes an array of groups anywhere
- technically it's possible to pass `only` and some route names, and those route names would then be ignored. we should probably validate the structure of the group config just a _bit_ to avoid this

bakerkretzmar:
Actually, decided to solve those issues by taking a slightly different approach, see #559.

PR #559, opened 2022-05-06
Add support for negating route filter patterns with `!`

bakerkretzmar:
Adds support for filtering _out_ certain route names by prefixing a pattern with `!`. If all provided patterns are negated, the patterns are treated as 'reject' rules, starting with all routes included by default and using the patterns to filter them out, which is how `except` works now.

Addresses #520 and #555 because groups now support both including and excluding routes. The existing functionality behaves like `only`, so the key addition would be a way to exclude routes, which this PR accomplishes:

```php
// Current usage, behaves like `only`
config(['ziggy.groups' => [
    'viewer' => [
        'posts.*',
    ],
]]);

// Added in this PR, behaves like `except`
config(['ziggy.groups' => [
    'guest' => [
        '!posts.*',
    ],
]]);

// Also added in this PR—mix and match! Behaves like `only` but additionally excludes negated patterns
config(['ziggy.groups' => [
    'viewer' => [
        'posts.*',
        '!posts.comments.*',
    ],
]]);
```

This does make the top-level `except` key a bit redundant, but that lines up nicely with my plan in v2 to combine 'only', 'except', and 'groups' into one feature 👀 

@YeeJiaWei thoughts? Your original example in the discussion would now look like this:

```php
return [
    'groups' => [
        'inertia' => ['!_debugbar.*', '!debugbar.*', '!ignition.*', '!admin.*'],
    ],
];
```

Closes #555.

### DIFF
--- a/src/Ziggy.php
+++ b/src/Ziggy.php
@@ -19,13 +19,13 @@ class Ziggy implements JsonSerializable
     protected $group;
     protected $routes;
 
-    public function __construct($group = null, string $url = null)
+    public function __construct($group = NULL, string $url = NULL)
     {
         $this->group = $group;
 
         $this->url = rtrim($url ?? url('/'), '/');
 
-        if (! static::$cache) {
+        if (!static::$cache) {
             static::$cache = $this->nameKeyedRoutes();
         }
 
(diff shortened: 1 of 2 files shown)


YeeJiaWei:
@bakerkretzmar I think this is also acceptable approach

PR #587, opened 2022-10-21
chore: Update README.md

afolabiabass:
This updates the guide on how to fetch route from json response when api call is made in frontend using fetch. See https://developer.mozilla.org/en-US/docs/Web/API/Response/json for more information


bakerkretzmar:
Whoops, thanks!

PR #600, opened 2022-11-27
Fix at script output format

MuriloChianfa:
Fixing argument requirement at script output format described in [issue#598](https://github.com/tighten/ziggy/issues/598#issue-1465484350)


bakerkretzmar:
See https://github.com/tighten/ziggy/issues/598#issuecomment-1328306506 – that `$function` argument is required for the script output formatter.

PR #630, opened 2023-04-20
Override route key name if $primaryKey is overridden

Jared0430:
Currently, if you use a custom primary key on your model, you also have to override the `getRouteKeyName()` method, otherwise, the custom route key name doesn't get detected

This PR fixes this by also checking if the `$primaryKey` property has been set on your model

### DIFF
--- a/tests/Unit/RouteModelBindingTest.php
+++ b/tests/Unit/RouteModelBindingTest.php
@@ -32,6 +32,12 @@ class RouteModelBindingTest extends TestCase
         $router->post('users', function (User $user) {
             return '';
         })->name('users.store');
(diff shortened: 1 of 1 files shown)


bakerkretzmar:
Thanks, good catch 👍🏻 can you add a test for this, and also add `getKeyName()` since we should probably be checking that too? I can circle back and add a test here in the next week or two as well, no rush.

Jared0430:
Added support for `getKeyName()` and some tests 👍

bakerkretzmar:
Thanks a lot! I'll tag a release in the next day or two.

PR #633, opened 2023-04-24
Get default path of output file from ziggy.output.path config

mhelaiwa:
This will help when working on multiple projects and using terminal aliases to generate the file. So instead of specifying the path in the command like `zgy {{output_path}}` it would be nice to just `zgy` and get the path according to the current project


bakerkretzmar:
Thanks!

PR #650, opened 2023-07-06
Allow setting URL in config

dshafik:
A very simply fix that allows setting the URL in the `config/ziggy.php` file like so:

```php
<?php
return [
    'url' => env('APP_URL'),
];
```

### DIFF
--- a/src/BladeRouteGenerator.php
+++ b/src/BladeRouteGenerator.php
@@ -11,7 +11,7 @@ class BladeRouteGenerator
 
     public function generate($group = null, $nonce = null)
     {
-        $ziggy = new Ziggy($group);
+        $ziggy = new Ziggy($group, config('ziggy.url', null));
 
         $nonce = $nonce ? ' nonce="' . $nonce . '"' : '';
 


bakerkretzmar:
What's your use case for this?

mikebrandl:
@bakerkretzmar Would imagine that perhaps @dshafik want's to compile the Javascript on a different server to where it's deployed.

We have this problem currently. Because it has the URL in the Ziggy.js file, this is still causing us some pain on the upgrade from 0.9.X

dshafik:
When using a CDN, or other proxy, the HOST header doesn't always match the public host name. 

We have an external (hosted by the CDN) and backend hostname (that is the public host at the cloud provider that is locked down so only the CDN hosts can access it) and then a further internal ONLY hostname that is not internet routeable. 

The host header seen by PHP is often not our public CDN one (usually it's the public cloud one), but that's the one Ziggy uses rather than the APP_URL env var.

bakerkretzmar:
@dshafik can you explain your specific setup in a bit more detail? I don't understand why your Laravel app isn't seeing the public URL of your actual site at all. Using some kind of proxy or load balancer is a common use case and Ziggy works with that setup as long as you've configured your [trusted proxies](https://laravel.com/docs/10.x/requests#configuring-trusted-proxies) and hosts properly.

Also, are you generating routes with the artisan command or using `@routes`?

bakerkretzmar:
Closing this for now, it's been explored before and didn't end up being necessary. @dshafik let me know if you aren't able to get this working using Laravel's [Trusted Proxies](https://laravel.com/docs/10.x/requests#configuring-trusted-proxies) and [Trusted Hosts](https://laravel.com/docs/10.x/requests#configuring-trusted-hosts), and if you can provide more details about your setup we can keep looking into this.

zymotik:
If you're using Ziggy with Vue and are looking to override Ziggy's build time config in an easy and simple way:

```ts
// app.ts
import { ZiggyVue } from 'ziggy-js'; // ./vendor/tightenco/ziggy/dist/vue.m
import { Ziggy } from './ziggy'; // from command output: @php artisan ziggy:generate ./resources/js/ziggy.js

Vue.use(ZiggyVue, { ...Ziggy, url: '' });
```

PR #660, opened 2023-08-17
Addition of Functionality to Skip Prefixes in Ziggy (mcamara/laravel-localization)

Rene-Roscher:
**Description:**

A new function has been introduced to skip certain prefixes in the Ziggy class.

**Background:**

The primary motivation for this change is the integration with the **mcamara/laravel-localization** package in conjunction with laravel inertiajs. An issue was identified during the locale update where the URLs were not being updated correctly, specifically, the prefix in the URI of this object was not set appropriately.

**Solution:**

With the new skipPrefix function, specific prefixes, such as the locale, can be skipped. This is particularly useful to ensure that URLs are generated correctly without the locale prefix. **By skipping the locale prefix, the mcamara/laravel-localization package defaults back to the previous locale** (provided it's configured to do so, e.g., via session).

This change ensures that URLs are generated and managed correctly, especially in applications that utilize both mcamara/laravel-localization and laravel inertiajs.


bakerkretzmar:
> An issue was identified during the locale update where the URLs were not being updated correctly, specifically, the prefix in the URI of this object was not set appropriately.

What exactly was the issue you identified? How were the URLs updated and how did you expect them to be updated? What prefix are you talking about, and what object do you mean?

Can you share some examples of what this PR does and what problem it solves? Are you able to use Ziggy's existing route filtering to accomplish this? I'm not really interested in adding functionality to integrate with or work around the implementations of specific other packages.

Please also ensure this branch is up to date with the latest code in the repo, it looks like your changes in Ziggy.php revert a recent PR with an unrelated bugfix.

Rene-Roscher:
Hello, sure I'll explain briefly.

I used the mcamara/laravel-localization package. The package is used for the localization in routing. So /en, /de and that is set very basic as prefix.

In the frontend I used InertiaJS & xiCO2k/laravel-vue-i18n for translated strings. 

Now when doing Inertia(router).replace(newUrl), ziggy is not updated because the source is not updated fundamentally (JS Script in blade loaded).

Now I noticed that you can also just omit the prefix since the package stores the language and that's when this came to me as a solution. 

You have in the config prefixes (e.g. /de, /en ...), which are then removed from the Ziggy routing, so that these prefixes are no longer in the URI - Since after a locale update (en => de) the routes with the old locale in the prefix were still (Say he was always sent back to the previous locale). 

A refresh of the page was out of question for me.

I tested the procedure and it should not cause any complications.

bakerkretzmar:
There are two ways you could accomplish this right now: route filtering or just reloading all the routes when the locale changes.

If that package creates separate prefixed routes for each locale, as long as the route _names_ are prefixed in addition to the paths, you should be able to filter them out by setting your `ziggy.exclude` config to something like `['en.*', 'de.*']`.

If you want to reload the routes without reloading the page, you can do that already by passing the Ziggy config down as Inertia shared data. Laravel Breeze does this out of the box: https://github.com/laravel/breeze/blob/1.x/stubs/inertia-common/app/Http/Middleware/HandleInertiaRequests.php#L38C34-L38C34

That second option gives you total control over loading the routes, so if filtering doesn't work with that package then that's the way to go.

If you have any other questions feel free to comment again, and the more actual concrete code examples you can share the easier it'll be to figure something out.

PR #661, opened 2023-08-17
Make slash "not encoding" less restrictive

aguingand:
### The problem
Following up on #500. This feature is great but too restrictive:
- The [docs](https://laravel.com/docs/10.x/routing#parameters-encoded-forward-slashes) state the "Encoded forward slashes are only supported within the last route segment." but in reality Laravel route generator doesn't encode slashes by default for all params (cf. [RouteUrlGenerator](https://github.com/laravel/framework/blob/85ad68cb0bcb5143e0e915098097c6d23ea22e7e/src/Illuminate/Routing/RouteUrlGenerator.php#L37))
- The `where` regex can be different than `.*`

Example:

```php
Route::get('/{breadcrumb}/page/{page}')->where('breadcrumb', '.+');
```
### Proposed solution
I move the `/` replacement after asserting the current param matches the given `wheres`. At this point if slashes are replaced it's because the RegExp allowed so.

### DIFF
--- a/src/js/Route.js
+++ b/src/js/Route.js
@@ -81,7 +81,7 @@ export default class Route {
         const [location, query] = url.replace(/^\w+:\/\//, '').split('?');
 
         const matches = new RegExp(`^${pattern}/?$`).exec(location);
-        
+
         if (matches) {
             for (const k in matches.groups) {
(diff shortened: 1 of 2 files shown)


aguingand:
I have tested in my project, it works well with a single starting param containing slash (both laravel + ziggy). But some js tests may be relevant here.

bakerkretzmar:
Thanks! Gonna come back to the encoding thing separately asap.

PR #667, opened 2023-09-15
Fix empty root URL with `absolute: false`

### LINKED ISSUE #666: Route `/` doesn't work with absolute=false
Hi!
Because of the removal of last `/`,  the home (`/`) route link doesn't work with Ziggy.
An update in the regexp should do the work. (I can do it if you want)
https://github.com/tighten/ziggy/blob/f2b43c03e49b0f717168b1d578c8c4f8c423daca/src/js/Route.js#L119-L119

Thanks,
Matt'

bakerkretzmar:
Fixes #666. Wanted to use a Regex lookbehind but support for them was only added to Safari several months ago and that's cutting it a bit close I think. That being said, the strings Ziggy works with are so short that we probably don't need to worry about performance at all, so we could probably lean on advanced Regex functionality more...

### DIFF
--- a/src/js/Route.js
+++ b/src/js/Route.js
@@ -26,7 +26,9 @@ export default class Route {
      * @return {String} Route template.
      */
     get template() {
-        return `${this.origin}/${this.definition.uri}`.replace(/(?<=.)\/+$/, '');
+        const template = `${this.origin}/${this.definition.uri}`.replace(/\/+$/, '');
+
+        return template === '' ? '/' : template;
     }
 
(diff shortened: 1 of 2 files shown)


Tofandel:
@bakerkretzmar I think instead of the lookbehind you could have just applied the regex only to `this.definition.uri`

bakerkretzmar:
Hmm yeah that's true, although that'll leave a trailing slash on a 'home' route with no path, if there is one.

Issue #680, opened 2023-10-18
Improvements on ziggy.d.ts generation


lmeysel:
I tried the newly introduced TS support and it works great so far.

I would like to discuss two further improvements (it is just cosmetics):

When generating the .d.ts file using `php artisan ziggy:generate <path>`, then `<path>` must point to a JS file. This made sense before the TS introduction, but when using `--types-only` flag, that is a bit inconvinient. Instead I suggest to allow arbitrary files and change the extension if it is not already .js/.d.ts respectively.

Second: In the readme, you ask to add another .d.ts file to basically shim the `routeFn` and make it globally available for TS. Since we also need generating the declarations for the route, it probably make sense, to add the shim to the definition (or at least having another flag in the command which creates the shim right along in the ziggy.d.ts)

bakerkretzmar:
That file path convenience makes sense 👍🏻 

The declarations I'm interested in but I have some questions. Just thinking out loud: In the example in the Readme we `import routeFn from 'ziggy-js';`. If we have to generate this declaration automatically we can't necessarily know that `ziggy-js` exists to import from—the user might not have the NPM package installed and might not have a bundler/TS alias set up to point `ziggy-js` to their vendor folder. (Actually, is that true? Or to get correct types without the NPM package would they _have_ to set up the path alias in their tsconfig, and therefore it would work fine?) That means we'd have to import from the vendor directory ourselves using a relative file path, which means we'll have to generate that path based on where the file is generated... which feels gross. Am I missing anything? I'll play around with this some more and see if I can answer some of these questions myself.

lmeysel:
I'll read this again tomorrow, but if I understand correctly: When generating the ziggy.d.ts you definitely will have the composer package installed. So from my perspective there is nothing wrong referencing the types from within the composer package (as you suggest to do in the tsconfig.json otherwise).

Maybe it is a good idea to give a bit more control about this to the user, e.g. like passing a flag (--ziggy-types=node|php). Alternatively (or additionally) you may inspect the package.json in CWD in order to find out there is the ziggy-js package installed. Or check the node_modules directory directly. When checking node_modules folder be careful of how you do this check, since diffrent package managers do different things (e.g. like not placing a folder but a symlink)

PR #687, opened 2023-11-02
Handle generated file extensions more robustly

bakerkretzmar:
This PR allows the `ziggy:generate` command to handle a wider range of `path` argument formats and still create the right files with the right extensions.

Before this PR, the path argument had to be a file name ending in `.js`. That makes sense if you're generating Ziggy's JavaScript config file, but not so much if you're generating its types too, and even less so if you're generating _just_ its types on their own.

Now, the extension of the path argument is mostly ignored and we just use the argument to figure out the file path and name. Path arguments ending in `.js`, `.d.ts`, `.ts`, or even no extension, all generate the requested `.js` and/or `.d.ts` file at the base filename specified. As an added convenience, if the path argument happens to point to an existing directory, `ziggy.js` and/or `ziggy.d.ts` files will be generated inside that directory instead.

This all applies to output paths set via config instead of arguments too.

See #680.

@lmeysel thoughts?

### DIFF
--- a/src/CommandRouteGenerator.php
+++ b/src/CommandRouteGenerator.php
@@ -33,20 +33,26 @@ class CommandRouteGenerator extends Command
     {
         $ziggy = new Ziggy($this->option('group'), $this->option('url') ? url($this->option('url')) : null);
 
-        $this->makeDirectory(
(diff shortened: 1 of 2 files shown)


lmeysel:
Looks good, nothing to add 🚀

PR #692, opened 2023-11-30
feat: generate types that match generated js

rrmesquita:
Right now, the README already asks the user to add a piece of typing in a `.d.ts` file in order to make the IDE aware of the global `route` function. Since this function is global and always available, I don't see a reason to leave that step to the user.

Another problem that this PR solves, is the error message shown when you want to import the Ziggy config from the generated JS file into a TS file.

```ts
import { Ziggy } from './ziggy'
      // ˆ Module '"./ziggy"' has no exported member 'Ziggy'. (ts(2305))
```

I'll keep this as a draft for now, as this change would also require changes to tests and the README. Let me know your thoughts, and thanks for creating this awesome lib that helps me a lot!


bakerkretzmar:
Very interested in making this work but need to think it through some more. We can't always count on being able to import from `ziggy-js` (see #680).

bakerkretzmar:
Thinking out loud—could this be an option for the `ziggy:generate` command? Like we include this declaration by default but add something like `--no-declare` to let people omit it if they need to?

rrmesquita:
If the `--no-declare` option is passed, it outputs the same as it does now, right? That could work I guess, but...

I see two types of people that would want these type definitions:
- They have a build step, so chances are that this person already have the npm package installed.
- They don't have a build step (maybe TALL?), uses the `route` helper injected by the `@routes` directive.

How problematic it would be to just require the user to install the npm package – which is super lightweight – at least as a dev dependency? If you think about, it makes sense, it fixes the issue, and it saves us from having to implement a not so elegant solution.

Let me know your thought and if there's another use-case I'm missing.

bakerkretzmar:
Definitely not going to require installing the npm package, the goal is for almost everyone to never need it.

neochief:
Hi!

From what I can figure out, the TS user of this package is required to set up their 'ziggy-js' dependency in some way anyway (either via tsconfig, or by having a NPM module). This is written in the TS section of the docs. So, I'm not sure what is the problem of importing from 'ziggy-js' in the TS context — it should be available in one way or the other. Besides, the current generated d.ts gives glaring TS error anyway.

I have tested the code suggested here, and it works fine. How about we proceed with this fix?

I should point out that the import in the suggested code should probably be changed to this one:

```
import { route as routeFn, type Config } from 'ziggy-js'
```

Related: https://github.com/tighten/ziggy/discussions/702

bakerkretzmar:
After reviewing this again I don't think I want to add the global declaration. We can't actually be _positive_ that users of the package have made their `route()` function available globally, and I'd rather make them add another couple lines of code elsewhere to make TypeScript aware of it than declare it for them and potentially silence real issues if they haven't imported or registered things correctly.

I'm also not clear what the `export const` stuff is doing—it shouldn't be a type error to import a variable from a regular JavaScript file, right? If `import { Ziggy } from './ziggy'` is importing from the generated JavaScript config file, not the file with the type declarations, I don't understand where the `no exported member` error is coming from. Feels like I'm missing something here.

Thanks a lot for the PR, I appreciate the work and the discussion! Going to close it for now but I'm definitely still open to trying to make the typing setup easier in the future.

Finally, if you want the file generated exactly like this in your own project, you can set up a custom formatter for the generate command and use that instead of our built in TypeScript one. See [the tests](https://github.com/tighten/ziggy/blob/5dc7a0e9e351bbea70852f65350781c07b4ab5ba/tests/Unit/CommandRouteGeneratorTest.php#L99) for an example and let me know if you have any questions.

PR #698, opened 2023-12-22
Use fresh routes from HandleInertiaRequest

guilheb:
As discussed in https://github.com/tighten/ziggy/discussions/696, it would be really nice for the `route()` function to always have up-to-date routes.

The simplest use-case is having a small set of routes for the whole world to see (login, registration, etc) and then have access to all routes once a user is logged in. This is easily achievable in HandleInertiaRequest, but the JavaScript object is never updated. This PR ensure the `route()` function uses the fresh list of routes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "guilheb/ziggy",
+    "name": "tightenco/ziggy",
     "description": "Generates a Blade directive exporting all of your named Laravel routes. Also provides a nice route() helper function in JavaScript.",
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
Thanks for the PR 👍🏻 Ziggy can't call `usePage()` internally because it's not just used in Inertia apps. The best way to make sure all routes are always available is to use the `@routes` Blade directive. If you don't want to do that the simplest alternative is overriding the route function in your own app, like you did.

PR #710, opened 2024-02-18
file cache route creation Ziggy.php

msoler75:
file cache creation of data for optimum performance. 

In my localhost, with a laravel project with ~300 routes, I reduce time about 10 to 56 ms

### DIFF
--- a/src/Ziggy.php
+++ b/src/Ziggy.php
@@ -26,8 +26,25 @@ class Ziggy implements JsonSerializable
 
         $this->url = rtrim($url ?? url('/'), '/');
 
-        if (! static::$cache) {
-            static::$cache = $this->nameKeyedRoutes();
+        if (!static::$cache) {
+            // el archivo ziggy se guarda en cache, aquí se comprueba si debe reconstruirse
+            $cache_routes = base_path("bootstrap/cache/routes-v7.php");
+            $cache_ziggy = base_path("bootstrap/cache/ziggy2.json");
+            if (
+                !file_exists($cache_ziggy) ||
+                !file_exists($cache_routes) ||
+                filemtime($cache_routes) > filemtime($cache_ziggy)
+            ) {
+                static::$cache = $this->nameKeyedRoutes();
+                file_put_contents($cache_ziggy, static::$cache->toJson());
+            } else {
+                try {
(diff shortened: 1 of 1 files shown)


bakerkretzmar:
Not totally convinced that this is necessary, but open to it if the performance gains are significant. I'm surprised that reading from the file and calling `collect()` is faster than doing it in memory.

Can you add tests for this and update your code comments to be in English? Also, is this tied to the route cache at all? How does it get cleared/updated when routes change?

Thanks!

alhawaj-bh:
I conducted benchmarks for file cache route creation to evaluate performance improvements.
In our app with over 4000 routes, profiling revealed that nameKeyedRoutes() is the top function in terms of execution time.


### Results Before:

- CPU Time: 139ms
- Wall Time: 176ms
- Memory Usage: 5.77MB
- Peak Memory Usage: 6.40MB


### Results After:

- CPU Time: 52ms (62% improvement)
- Wall Time: 57ms (68% improvement)
- Memory Usage: 7.39MB
- Peak Memory Usage: 7.78MB



The changes significantly reduce CPU and wall time while slightly increasing memory usage.

profiler-019358ad-bf3e-732b-8b0d-05613f07caee
profiler-0193589e-fd30-705c-beca-0637e650059b (1)

bakerkretzmar:
Going to close this for now but thanks a lot for the PR!

msoler75:
I tested in production:

Standard ziggy constructor:
> php artisan app:profile-ziggy
10.95 ms

Constructor with json file cache:
> php artisan app:profile-ziggy
0.82 ms

My routes:
bootstrap/cache/routes-v7.php  size 904 kb
bootstrap/cache/ziggy2.json  size  54 kb  <-- this is the json file cache created


Code of console command profiler profileZiggy.php:

```
public function handle()
    {
        // we clear routes here
        new \Tightenco\Ziggy\Ziggy();
        \Tightenco\Ziggy\Ziggy::clearRoutes();

        // now we load constructor
        $start = microtime(true);
        $ziggy = new \Tightenco\Ziggy\Ziggy();
        $end = microtime(true);
        $time = ($end - $start) * 1000;
        $this->line(number_format($time, 2));
    }

```

PR #717, opened 2024-02-24
Allow for .ts routes file

Stnaire:
Small changes to allow the generate command to output .ts files.

### DIFF
--- a/src/CommandRouteGenerator.php
+++ b/src/CommandRouteGenerator.php
@@ -41,18 +41,18 @@ class CommandRouteGenerator extends Command
             $this->makeDirectory($path);
         }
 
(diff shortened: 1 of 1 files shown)


bakerkretzmar:
Thanks! This makes sense but it needs a bit more work. Just thinking out loud, I'm happy to make these changes.

- Need to add/update tests.
- Need to check a bit more carefully in the first path normalization because if they pass a `.d.ts` filename I think that'll make it through and be used by Ziggy to generate the file, which I don't think we want.
- I think the generated file may contain type errors, we might need something like this but I'm not sure and I don't know if there would be other side effects:

```ts
declare global {
    var Ziggy: Config;
}
```

bakerkretzmar:
Closing for now but not opposed to considering something like this in the future.

PR #721, opened 2024-02-29
Export the types file

simon-tma:
If there is an `exports` key set, Typescript requires the types file to be included in the list of the files exported.

Without this change, Typescript errors when trying to import types with:
```
error TS7016: Could not find a declaration file for module 'ziggy-js'. '/.../node_modules/ziggy-js/dist/index.js' implicitly has an 'any' type.
  There are types at '/.../node_modules/ziggy-js/src/js/index.d.ts', but this result could not be resolved when respecting package.json "exports". The 'ziggy-js' library may need to update its package.json or typings.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,10 @@
     ],
     "type": "module",
     "source": "./src/js/index.js",
-    "exports": "./dist/index.js",
+    "exports": {
+        "default": "./dist/index.js",
+        "types": "./src/js/index.d.ts"
+    },
     "module": "./dist/index.esm.js",
     "types": "./src/js/index.d.ts",
     "scripts": {
--- a/package.json
+++ b/package.json
@@ -35,8 +35,8 @@
     "type": "module",
     "source": "./src/js/index.js",
     "exports": {
-        "default": "./dist/index.js",
-        "types": "./src/js/index.d.ts"
+        "types": "./src/js/index.d.ts",
+        "default": "./dist/index.js"
     },
     "module": "./dist/index.esm.js",
     "types": "./src/js/index.d.ts",


Mohammad-Alavi:
First of all thanks for this great package ❤️ . I really mean it.
Also, your docs are the worst I have ever seen ❤️  (sorry, just wanted to get it out of chest!)

Please merge this one. This is really a high priority fix.

bakerkretzmar:
@simon-tma what version of TypeScript are you using? Can you link me to where TypeScript says `types` is **required** (I can see that it's available, but it looks optional). Happy to add this if it's necessary but the TS docs make it seem like everything should work fine either way.

simon-tma:
@bakerkretzmar  After further digging (using https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution, https://github.com/microsoft/TypeScript/issues/52363 and https://arethetypeswrong.github.io/?p=ziggy-js%402.0.4), it's not that `types` is required in `exports`, it's that the top-level one is ignored when doing resolution using `exports` so typescript falls back to looking for sibling files instead.

Resolution trace from "Are The Types Wrong" showing what paths are being searched for the declaration file:
Screenshot 2024-03-20 at 08 57 25

I'm seeing this with TS 5.3 and 5.4, though it's likely to happen with any version of Typescript using `nodenext` or `node16` for module resolution.

bakerkretzmar:
@simon-tma gotcha, makes sense. Appreciate you sharing those details, and that "are the types wrong" tool is very cool.

bakerkretzmar:
Released just now in v2.0.5. Thanks again for digging into this 🙏🏻

PR #722, opened 2024-03-04
Don't write to the output files if they have not changed at all

gms8994:
When trying to include the `ziggy:generate` run as part of `webpack` (via `webpack-watch-files-plugin` and `hook-shell-script-webpack-plugin`), `ziggy:generate` will always write the output files even if there are no changes to the source routes file(s), which will then cause `webpack` to see a change to some of the source files that are being watched, which will trigger `ziggy:generate` to run again, causing an infinite loop.

### DIFF
--- a/src/CommandRouteGenerator.php
+++ b/src/CommandRouteGenerator.php
@@ -46,13 +46,19 @@ class CommandRouteGenerator extends Command
         if (! $this->option('types-only')) {
             $output = config('ziggy.output.file', File::class);
 
-            $this->files->put(base_path("{$name}.js"), new $output($ziggy));
+            $content = new $output($ziggy);
+            if ($this->files->get(base_path("${name}.js")) !== (string)$content) {
+                $this->files->put(base_path("{$name}.js"), $content);
+            }
         }
 
         if ($this->option('types') || $this->option('types-only')) {
             $types = config('ziggy.output.types', Types::class);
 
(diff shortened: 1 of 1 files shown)


bakerkretzmar:
Thanks for the PR! Can you configure your watcher to only run `php artisan ziggy:generate` if one of your route files is changed/saved, not on every Webpack run? Ideally instead of doing a check like this within Ziggy, the command would just only run when it actually needs to.

bakerkretzmar:
Closing for now, looks like this should be possible within your Webpack config, which I think is the right place for this kind of thing to be handled. Here's a simplified plugin I have working locally that you can adapt to your project:

```js
class Plugin {
    apply(compiler) {
        compiler.hooks.watchRun.tap('ziggy-generate', function (compiler) {
            if (compiler.modifiedFiles && Array.from(compiler.modifiedFiles).find((file) => file.includes('/routes/'))) {
                console.log('Route file changed, regenerating Ziggy config:')
                execSync('php artisan ziggy:generate');
            }
        });
    }
}
```

This should cause changes to your route files to trigger 2 builds, one because of the change and one from the generate command updating the `ziggy.js` file, but then it should stop.

Let me know if there's anything else I can help with here!

gms8994:
@bakerkretzmar given this, do you think it would be possible to include your plugin class with the ziggy code so that it could easily be included in the webpack config? You and I can't be the only ones that want this kind of functionality :)

Perhaps it could take params for the paths to watch and the command to run?

PR #743, opened 2024-04-09
fix route().params type

peter-emad99:
As mentioned in docs route().params will always return the params as string value


bakerkretzmar:
I can't tell what actually changed here, can you revert your formatting changes so the diff only contains the fix you're suggesting? Thanks.

peter-emad99:
sorry for confusion .. just changed the route params type from unknow to string

bakerkretzmar:
Thanks!

PR #744, opened 2024-04-10
add ability to acess only the query params or the actual route params

### LINKED ISSUE #742: Feature: Add ```route().queryParams``` to get only query params without laravel route params
#### Description
currently route().params method get all route params .. including the laravel route params 
in some situations, we need to access only route query params  

#### Alternatives

the only way to get do this now with Browser Url or UrlSearchParams api but it's bit different ..because route() use qs under the hood that has different params parsing mechanism

peter-emad99:
Added two new methods to Router

`route().queryParams` to get only the query paramters parsed with the qs
`route().routeParams` to get only the actual route params without query params

i didn't like the routeParams name as it's no clear enough but i couldn't find another name 

this complete #742 

i didn't update the docs to inlude this but if approved will do that


bakerkretzmar:
What's your use case for this? Are you primarily interested in the route params or the query params?

I'm on the fence about this, because there are other ways to get most of this info already, like with URLSearchParams, and because Laravel's `route()` helper doesn't do this directly.

peter-emad99:
as i mentioned in the issue ... using UrlSearchParams don't parse the params the same way when using the route() function because of it using qs under the hood ..
i have some cases  i wanted to access only the route params or the query params

when there is a route parameter and query parameter with the same name ..and you want to access only the query or route... with route().params the one from the query will always overwrite the route params  ```{...params,...query}``` 

it gives more control to apply filters  and search  and apply reset  with something like this 
```router.reload(route(route.().current(),route().routeParams))``` or
 ```router.reload(route(route.().current(),{...route().params,...filterData}))``` ,

yes we can achieve the same thing with some little js but why when we have a better and easy way using route() herlper ?

PR #747, opened 2024-04-15
Ziggy Import Vue

alnutile:
Maybe I am missing a step? This is the only way I can get the import to work without getting an error from vite.

Laravel 10 and 11.
"tightenco/ziggy": "^2.1"

Hope this helps?

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ Ziggy includes a Vue plugin to make it easy to use the `route()` helper througho
 
 ```js
 import { createApp } from 'vue';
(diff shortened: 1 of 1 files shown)


bakerkretzmar:
That particular example is based on the section right above, which shows how to add an alias in Vite to point `ziggy-js` to that vendor path: https://github.com/tighten/ziggy?tab=readme-ov-file#importing-the-route-function. Importing from `../../vendor/tightenco/ziggy` is fine too though.

PR #760, opened 2024-06-08
fix: infinite loop

M0einRahmani:
"I used (new Ziggy())->jsonSerialize() in Exceptions\Handler.php in my Laravel + InertiaJS project. When an error or exception occurs, I noticed that the resolveBindings function gets stuck in an infinite loop."


bakerkretzmar:
Can you explain what exactly the issue is here and how this PR solves it? Also please revert the formatting changes that aren't part of your fix.

PR #763, opened 2024-07-03
Laravel route change

hasanablak:
We already mention in the comment line: "// routes/api.php".

if we are say 
"Route::get('api/ziggy', fn () => response()->json(new Ziggy));"

this time laravel will create like this route:

"api/api/ziggy"


bakerkretzmar:
Haha yep you're totally right, great catch. Thanks!

PR #766, opened 2024-07-22
Escape special regex characters when matching routes

aguingand:
I noticed routes with dots inside (e.g. `{post}.{slug}`) does not match using `route().current()`. Indeed RegEx special characters are currently not escaped, so I added a remplacement for :
```
.*+^$()|[]
```
I excluded `{}?` chars because there are needed for parameters.

### DIFF
--- a/src/js/Route.js
+++ b/src/js/Route.js
@@ -80,6 +80,7 @@ export default class Route {
         // Transform the route's template into a regex that will match a hydrated URL,
         // by replacing its parameter segments with matchers for parameter values
         const pattern = this.template
+            .replace(/[.*+^$()|[\]]/g, '\\$&')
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
What is your replacement actually doing exactly? And what's your use case for this?

bakerkretzmar:
Oh never mind I get it. Thanks! This makes sense, I'm gonna take a look at how Laravel handles routes like this because I'm fine erroring on characters that break the actual route. Not sure if it makes a difference but I'm pretty sure `^` and `|` aren't valid in URLs anyway unless they're encoded.

aguingand:
Sorry I've rushed the description a little bit ! Yes the goal is to backlash these chars to be matched in the regex. 
You're right `^` & `|` can be removed as there are encoded. Laravel seems to be ok with all these chars in a route.

PR #770, opened 2024-08-13
Optimize has()

thomas-brx:
O(n) -> O(1) for route checks.

In our application, the `has` method came out on the top in our profiling. It may be that we are calling it too often for some reason, but it prompted this PR to optimize the method.

Note - got the tests running after patching one FolioTest to ignore ordering of the routes array. Don't know if this is ok or not.

### DIFF
--- a/src/js/Router.js
+++ b/src/js/Router.js
@@ -211,7 +211,7 @@ export default class Router extends String {
      * @return {Boolean}
      */
     has(name) {
-        return Object.keys(this._config.routes).includes(name);
+        return this._config.routes.hasOwnProperty(name);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
Thanks! Just for reference (I had to look it up), O(n)→O(1) means that currently this operation's performance depends on the size of the routes object, so with thousands of routes it would take longer to make a `has()` check, whereas after this PR that won't matter, checking for a route will take the same amount of time no matter how many routes there are—right?

thomas-brx:
> Thanks! Just for reference (I had to look it up), O(n)→O(1) means that currently this operation's performance depends on the size of the routes object, so with thousands of routes it would take longer to make a `has()` check, whereas after this PR that won't matter, checking for a route will take the same amount of time no matter how many routes there are—right?

Thanks @bakerkretzmar, that is a perfect explanation :) Sorry for being so cryptic in my description.
Your work with ziggy is much appreciated 🏅

PR #771, opened 2024-08-13
Optimize has()

thomas-brx:
O(n) -> O(1) for route lookups.

Same patch as #770 for the 1.x branch.


bakerkretzmar:
Thanks, but we're not updating 1.x anymore.

PR #780, opened 2024-10-20
Add optional gracefull error handling

mhelaiwa:
Add optional gracefull error handling through `config/ziggy.php`.  This will enable the user to receive an error through the console without stopping app's execution and fallback to a specified url like `#`

```
<?php
// config/ziggy.php
return [
    ... 
    'errors' => [
        'graceful' => true,
        'fallback_url' => '#zzzzzz'
    ]
   ...
];
```


bakerkretzmar:
Thanks, really appreciate the PR. Going to decline to add this functionality at the moment for two main reasons. First, Laravel doesn't do this, and one of our main priorities with Ziggy is to match Laravel's `route()` helper behaviour as closely as possible (in this case that means erroring like Laravel does). Second, I don't think we'd actually want to 'fall back' to a different route in the case that Ziggy is passed a route that doesn't exist, because that shouldn't happen—if you're passing a missing route that's a bug and it should be corrected, or ideally caught and removed during testing, it's not a real world use case we need to accommodate. Thanks again!

PR #785, opened 2024-11-03
Add note to use inject for vue 3

mhelaiwa:
Add note to use inject for vue 3 to be able to access route() function
Related to issue https://github.com/tighten/ziggy/issues/768


bakerkretzmar:
Thanks!

PR #786, opened 2024-11-06
Fix type definition for route() with only options

bram-pkg:
The order of function definitions was disallowing `route(undefined, undefined, undefined, config)` to be matched. This PR fixes that.

I've also added a test to ensure this doesn't fail again in the future.


bakerkretzmar:
Thanks! I could have sworn I put them in the previous order specifically to avoid an issue like this where one of the definitions wasn't getting matched, but clearly your fix here works.

bram-pkg:
Thanks!

PR #790, opened 2024-11-19
Only limit route names when calling route()

simon-tma:
When `strictRouteNames` is set, we still need to allow wildcard route names to be passed to `current()` and potentially invalid route names to be passed to `has()`.

This changes allows `route().current('dashboard.*')` to be considered valid while still preventing calling `route()` with an unknown route name.


bakerkretzmar:
Great catch and appreciate your patience. Thanks for the PR!

PR #792, opened 2024-11-24
Fix deprecated implictly nullable parameter

datlechin:
Fix Deprecate implicitly nullable parameter types, see https://wiki.php.net/rfc/deprecate-implicitly-nullable-types

### DIFF
--- a/src/BladeRouteGenerator.php
+++ b/src/BladeRouteGenerator.php
@@ -9,7 +9,7 @@ class BladeRouteGenerator
 {
     public static $generated;
 
-    public function generate($group = null, string $nonce = null): string
+    public function generate(array|string|null $group = null, ?string $nonce = null): string
     {
         $ziggy = new Ziggy($group);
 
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
Can you remove the `array|string|null` type declaration you added in this PR? I don't believe that's actually an issue for the deprecation thing because that parameter isn't typed at all, and adding those types would be a backwards compatibility break (even if they're accurate).

src/BladeRouteGenerator.php
```diff
@@ -9,7 +9,7 @@ class BladeRouteGenerator
 {
     public static $generated;
 
-    public function generate($group = null, string $nonce = null): string
+    public function generate(array|string|null $group = null, ?string $nonce = null): string
```
datlechin:
```suggestion
    public function generate($group = null, ?string $nonce = null): string
```

src/Ziggy.php
```diff
@@ -24,7 +24,7 @@ class Ziggy implements JsonSerializable
     protected Collection $routes;
 
     public function __construct(
-        protected $group = null,
+        protected array|string|null $group = null,
```
datlechin:
```suggestion
        protected $group = null,
```

bakerkretzmar:
Thanks.

PR #797, opened 2024-12-12
feat(vue): allow to change global/inject route name

### LINKED ISSUE #796: [vue] Allow to change `route` global name via options.
#### Description

In Vue, it is usually advised to prepend the global mixin names with a `$`.
At the moment, if someone wants to import `route` and use it in the template, the global mixin types will take precedence.

My use case is that I want to deprecate the global usage with a `@deprecated` JSDoc attribute, and use the local import, but I still see the deprecated from global.

Set `route` TS definition.

image

Use `route` import in template

image

It would be great if we could provide what name to be used for the `route` helper, and default to `route` if not defined.

image

#### Suggestion

Add `routeName` to the `ZiggyVue` plugin options.

#### Alternatives

The alternative is to manually add the `route` mixin under a different name.

image

Type it accordingly

image

No more conflicts with the import

image

unrevised6419:
Closes #796

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,6 +388,17 @@ const route = inject('route');
 </script>
 ```
 
+With `<script setup lang="ts">` you can use TypeScript to type the `route` function:
+
+```vue
+<script setup>
+import { inject } from 'vue';
+import { route as routeFn } from 'ziggy-js'
+
+const route = inject<typeof routeFn>('route')
+</script>
+```
+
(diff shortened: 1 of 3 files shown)


bakerkretzmar:
Thanks a lot for the PR and the detailed issue! We're going to pass on adding this to the plugin for now, it's really just a convenience and we want to keep it super simple. If you want to rename the helper you should create the mixin/provider yourself just like you demonstrated in #796, instead of using Ziggy's included Vue plugin. That way is also less code and gives you complete control over the implementation.

PR #800, opened 2024-12-23
Add some type hinting to fix a warning about implicitly marking param

bendauphinee:
Spotted this warning on my dev console, so I've tweaked the type handling a little bit.

│ Tighten\Ziggy\BladeRouteGenerator::generate(): Implicitly marking parameter... │

### DIFF
--- a/src/BladeRouteGenerator.php
+++ b/src/BladeRouteGenerator.php
@@ -9,7 +9,7 @@ class BladeRouteGenerator
 {
     public static $generated;
 
-    public function generate($group = null, ?string $nonce = null): string
+    public function generate(?array $group = null, ?string $nonce = null): string
     {
         $ziggy = new Ziggy($group);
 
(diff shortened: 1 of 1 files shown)


src/BladeRouteGenerator.php
```diff
@@ -28,10 +28,10 @@ public function generate($group = null, ?string $nonce = null): string
         return (string) new $output($ziggy, $routeFunction, $nonce);
     }
 
-    private function generateMergeJavascript(Ziggy $ziggy, string $nonce)
+    private function generateMergeJavascript(Ziggy $ziggy, string $nonce): string
     {
         $output = config('ziggy.output.merge_script', MergeScript::class);
-
+dd($output);
```
efureev: del

bendauphinee: Resolved, of course I left that in lol.

bendauphinee:
Apparently I have to resolve some test errors. I'll flip this PR back to ready once I have.

bakerkretzmar:
Pretty sure if we want to type the `$group` parameter it'll have to be `mixed` or `?mixed` for backwards compatibility, but at the very least it needs to accept strings too because when it's passed it's usually a string.

PR #805, opened 2025-01-19
feat: support passing except/only args to ziggy command

joshmanders:
This PR adds the ability to pass `--except` and `--only` args to the artisan command.

The reason for this is I have developed a [vite plugin](https://github.com/aniftyco/vite-plugin-ziggy) for auto generating of your ziggy routes and types when any changes happen to `routes/` files and it would be nice to be able to just `composer require tightenco/ziggy` then `npm i -D vite-plugin-ziggy` and add the plugin to your vite config and it just works.

I have this setup to respect config as usual, but if you pass one of the args to the command that takes precidence over the same key in the config.

Only supports `except/only`, the rest should delegate to the config if you want to get that deep, in which case if you don't pass an `except` or `only` key to my plugin, it won't pass down to the command allowing ziggy to work as always.

### DIFF
--- a/README.md
+++ b/README.md
@@ -550,7 +550,7 @@ Route::get('ziggy', fn () => response()->json(new Ziggy));
 
 ### Re-generating the routes file when your app routes change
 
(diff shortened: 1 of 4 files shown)


bakerkretzmar:
Thanks! The Vite plugin looks cool, is storing generated files in `node_modules` a common practice for JS packages? I'd be worried about that breaking or causing other obscure issues but it sure is clean.

joshmanders:
> Thanks! The Vite plugin looks cool, is storing generated files in `node_modules` a common practice for JS packages? I'd be worried about that breaking or causing other obscure issues but it sure is clean.

@bakerkretzmar yeah a few other projects do it, you could even store them in `node_modules/.cache` as that's what it's intended for, instead of dumping things into project roots. The largest project I know of that does this and where I got it from is [Prisma ORM](https://prisma.io).

PR #809, opened 2025-01-31
Fix `current()` type in strict mode

### LINKED ISSUE #808: Typescript error with route().current() RouteName
### Ziggy version

v.2.5.0

### Laravel version

v11.39.1

### Description

Related to: https://github.com/tighten/ziggy/commit/5dc7a0e9e351bbea70852f65350781c07b4ab5ba

Since 2.5.0 I now get a typescript error with the below code when using strict mode: Type 'string & {}' is not assignable to type 'keyof RouteList'.

Suggested fix:

```
/**
 * Ziggy's Router class.
 */
interface Router {
    current(): RouteName | undefined; // Change this to ValidRouteName
    current<T extends RouteName>(name: T, params?: ParameterValue | RouteParams<T>): boolean;
    get params(): Record<string, string>;
    get routeParams(): Record<string, string>;
    get queryParams(): ParsedQs;
    has<T extends RouteName>(name: T): boolean;
}
```

Also I believe it would be valuable to export the ValidRouteName type definition (and perhaps some others) so it can be used for wrapper / helper code. In this situation I would be able to make a type assertion if the suggested fix is not viable for some reason.

### Ziggy call and context

```js
<Button
    href={route(route().current()!)} // Error here
>
    reload
</Button>
```

### Ziggy configuration

```js
//
```

### Route definition

```php
//
```

bakerkretzmar:
Updates `route().current()` and `route().has()` to use the new `ValidRouteName` type so they only return or accept known route names in strict mode. Technically `route().current()` should _always_ only return known route names, but I don't think we can actually use the `KnownRouteName` type because the interface it looks in for route names could be empty if the user hasn't run the command to generate it.

Fixes #808.

### DIFF
--- a/src/js/index.d.ts
+++ b/src/js/index.d.ts
@@ -170,12 +170,12 @@ interface ParsedQs {
  * Ziggy's Router class.
  */
 interface Router {
-    current(): RouteName | undefined;
-    current<T extends RouteName>(name: T, params?: ParameterValue | RouteParams<T>): boolean;
+    current(): ValidRouteName | undefined;
+    current<T extends ValidRouteName>(name: T, params?: ParameterValue | RouteParams<T>): boolean;
     get params(): Record<string, string>;
     get routeParams(): Record<string, string>;
     get queryParams(): ParsedQs;
-    has<T extends RouteName>(name: T): boolean;
+    has<T extends ValidRouteName>(name: T): boolean;
 }
 
 /**


simon-tma:
This has undone #790, thus `route().current('dashboard.*')` is now erroring again

bakerkretzmar:
Yikes thanks, will get that fixed.

PR #815, opened 2025-02-12
Clarify Router class behavior in README

travisricks:
As a follow-up to this discussion: https://github.com/tighten/ziggy/discussions/814, I'd like to propose adding a note to clarify the expected behavior of `route().current()` on filtered out routes.


bakerkretzmar:
Thanks for the PR! Going to pass on this for now.

PR #816, opened 2025-02-12
fix type hinting for PHP 8.4

bodny:
Fix of deprecation errors:
```
Implicitly marking parameter as nullable is deprecated, the explicit nullable type must be used instead
```

This is also attempt to fix PR #800 https://github.com/tighten/ziggy/pull/800, so these changes can be merged.


src/BladeRouteGenerator.php
```diff
@@ -9,7 +9,7 @@ class BladeRouteGenerator
 {
     public static $generated;
 
-    public function generate($group = null, ?string $nonce = null): string
+    public function generate(array|string|bool|null $group = null, ?string $nonce = null): string
```
bakerkretzmar:
```suggestion
    public function generate(array|string|null $group = null, ?string $nonce = null): string
```

src/Ziggy.php
```diff
@@ -24,7 +24,7 @@ class Ziggy implements JsonSerializable
     protected Collection $routes;
 
     public function __construct(
-        protected $group = null,
+        protected array|string|bool|null $group = null,
```
bakerkretzmar:
```suggestion
        protected array|string|null $group = null,
```

bakerkretzmar (changes requested):
Thanks! Just remove `bool`.

PR #821, opened 2025-02-27
Update extendability of `Types.php`

rrmesquita:
Customizing the types output is something I do in every project. This is just a simple QOL PR that saves me from having to port all that routes collection step.


bakerkretzmar:
Thanks!

PR #829, opened 2025-03-21
Update the exported member in Types.php

### LINKED ISSUE #828: TypeScript: `Module ... has no exported member Ziggy`
### Ziggy version

2.5.2

### Laravel version

12

### Description

I generate my routes and types. When I import the routes config in typescript the TS service throws an error:

> Module "@/lib/ziggy" has no exported member Ziggy

It makes sense because the `ziggy.d.ts` doesn't export anything except `export {};` (to make it a module).

### Ziggy call and context

```js
Not relevant
```

### Ziggy configuration

```js
Not relevant
```

### Route definition

```php
Not relevant
```

Livijn:
Fixes https://github.com/tighten/ziggy/issues/828 by exporting the correct Ziggy. Updated the tests as well.

### DIFF
--- a/src/Output/Types.php
+++ b/src/Output/Types.php
@@ -23,7 +23,7 @@ class Types implements Stringable
         declare module 'ziggy-js' {
           interface RouteList {$routes->toJson(JSON_PRETTY_PRINT)}
         }
-        export {};
+        export { Ziggy } from 'ziggy-js';
 
         JAVASCRIPT;
     }
--- a/tests/fixtures/ziggy.d.ts
+++ b/tests/fixtures/ziggy.d.ts
@@ -31,4 +31,4 @@ declare module 'ziggy-js' {
     ]
 }
 }
-export {};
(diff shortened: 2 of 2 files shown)


Spice-King:
This is incorrect, `ziggy-js` does not export a member called `Ziggy`. The correct way to do the types is: 
```ts
import { Config } from "ziggy-js";
export const Ziggy: Config;
```
And that's only if `ziggy:generate` not run with `--types-only`, as there is no generated JS file to import in that case.

I've personally just subclassed the `Output\Types` class, set Ziggy to use it, then overrode `__toString` to prepend that bit of TS. Given I have Vite auto-generating using `--types`, the `--types-only` flag is a non issue for me right this second.

bakerkretzmar:
Closing for now because @Spice-King is correct that `ziggy-js` does not actually export `Ziggy`. I am going to keep looking into this typing/naming issue though. Thanks anyway!

PR #846, opened 2025-05-31
Replace `qs`

benmccann:
https://npmgraph.js.org/?q=qs - 18 dependencies
https://npmgraph.js.org/?q=neoqs - 0 dependencies


bakerkretzmar:
Thanks! We're pinned on `qs` v6.9.7 which already has no dependencies and is smaller than `neoqs`, but I do want to move away from `qs` at some point so I'll look into this some more.

benmccann:
Oh, I didn't realize older versions were better in that regard. Thanks for sharing!

Here's a page describing several alternatives to `qs`: https://github.com/es-tooling/module-replacements/blob/main/docs/modules/qs.md

I chose `neoqs` as being the easiest to migrate to, but think any of the others would be fine too if you were to choose one of them.

bakerkretzmar:
Going with `qs-esm` purely because it's owned by a large organization and has way more installs, so I feel a bit better about its long term maintenance. Not married to it though. Thanks again for the prompt on this!

PR #854, opened 2025-07-05
Support route model binding and interfaces

christian98:


bakerkretzmar:
Thanks! Really appreciate the simple failing test, genuinely one of the best ways to start a PR 🙏 

Took me a while to figure out but in the end this turned out to be suspiciously simple. I'm glad it works now because the use case behind this, where your route parameter conforms to some TypeScript interface that is specific about its properties (like a model), is probably pretty common. I'm doing this more in my own projects too.

PR #857, opened 2025-07-31
Add Config class for Ziggy configuration management (dynamic configuration)

thaqebon:
#### Summary
* Introduced a `Config` class to manage Ziggy configuration.
* Added static `set()`, `get()` and `absolute()` methods for default configuration.
* Updated `Route` and `Router` classes to use the new `Config` class.
* Added tests to ensure the config behavior is consistent and reliable.

#### Purpose
The main goal of this PR is to allow **dynamic configuration** of Ziggy at runtime by setting a config getter:

```js
 // Inertia example
ZiggyConfig.set(() => usePage().props.ziggy);
 // DOM-based config
ZiggyConfig.set(() => JSON.parse(document.getElementById('ziggy-routes-json').textContent));

// Set default absolute flag
ZiggyConfig.absolute(false)
```

This makes Ziggy more flexible by allowing dynamic configuration instead of relying on a static one.

### DIFF
--- a/src/js/Config.js
+++ b/src/js/Config.js
@@ -180,17 +180,22 @@ export default class Config {
      * @return {Object}
      */
     get location() {
-        const {
-            host = '',
-            pathname = '',
-            search = '',
-        } = typeof window !== 'undefined' ? window.location : {};
+        const windowLocation = typeof window !== 'undefined' ? window.location : {};
+        let configLocation = this._config.location || {};
 
-        const location = this._config.location ?? {};
+        if (typeof configLocation === 'string') {
+            try {
+                const { host, pathname, search } = new URL(configLocation);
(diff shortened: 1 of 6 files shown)


bakerkretzmar:
Thanks a lot for the PR, and for your patience! This is really interesting but I think I'm going to pass for now. It's a lot of new code to introduce and I'm not sure that this pattern is super common or necessarily something we want to push people towards.

I can see how it could be helpful to eventually have something a bit more polished than `Ziggy.location = '...'` in Ziggy to handle updating configuration over time, especially for Inertia. If we do go down that road though I think I would want to lean towards something that updates it when it changes, something like `route().configure({ ... })` to pass in new options, as opposed to fetching it dynamically on every `route()` call. Not sure though.

PR #862, opened 2025-09-04
feat(typescript): add branded type for creating routes

pataar:
This adds a branded type for RouteUrl. This way you can create custom Anchor components, custom functions that are typed to a RouteUrl. 

e.g.

```ts
function redirect(route: RouteUrl){
   // redirect logic
}

const route = router('home');

if(condition){
  redirect(route);
  redirect("https://random.url"); // error
}
```


bakerkretzmar:
Very cool, thanks! I took out the export of `KnownRouteName` because I don't think it's necessary for this PR's functionality and I'm still very hesitant to expose Ziggy's types publicly in case we rename things or accidentally lock ourselves into something. If that was actually necessary for something here and I missed it please let me know.

pataar:
That's all good. Had some personal use cases for it. But I can extract it with Parameters<> as well.

Thanks for merging!

PR #875, opened 2026-03-05
Fix shared route bindings mutation in _substituteBindings

pataar:
When `_substituteBindings` encounters an object missing its registered binding key (e.g. `slug`), it falls back to `id` — but it does so by writing `bindings[key] = 'id'` directly on the shared route definition. This permanently overwrites the original binding for all subsequent calls.

For example, with a `posts.show` route bound to `slug`:

```js
// First call — partial object, no slug, falls back to id
route('posts.show', { post: { id: 1 } });
// 'https://example.com/posts/1' ✓

// Second call — full object with slug
route('posts.show', { post: { id: 1, slug: 'hello-world' } });
// Expected: 'https://example.com/posts/hello-world'
// Actual:   'https://example.com/posts/1' ✗
```

The second call uses `id` instead of `slug` because the first call mutated the route config.

This PR replaces the mutation with a local variable so the fallback is resolved per-call.

### DIFF
--- a/src/js/Router.js
+++ b/src/js/Router.js
@@ -321,18 +321,19 @@ export default class Router extends String {
                 return { ...result, [key]: value };
             }
 
-            if (!value.hasOwnProperty(bindings[key])) {
-                if (value.hasOwnProperty('id')) {
-                    // As a fallback, we still accept an 'id' key not explicitly registered as a binding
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
Wow great catch, thank you!

PR #878, opened 2026-03-06
Fix TypeError when optional params omitted from array (fixes #736)

### LINKED ISSUE #736: TypeError when optional params omitted from array
Given this route:

```js
const routes = {
    'posts.comments.show': {
        uri: 'posts/{post}/comments/{comment?}',
        methods: ['GET', 'HEAD'],
        parameters: ['post', 'comment'],
    },
};
```

Which generates these types:

```ts
interface RouteList {
    'posts.comments.show': [
        { name: 'post'; required: true },
        { name: 'comment'; required: false },
    ];
}
```

These valid `route()` calls, which **should not** cause any type errors, do:

```ts
route('posts.comments.show', [2]);
route('posts.comments.show', ['foo']);
route('posts.comments.show', [{ id: 2 }]);
```

The issue is that our type for an array of parameters doesn't specify any of them as optional (except additional unknown params after the ones we're expecting), so e.g. passing an array with less than 2 items will error if the route has 2 parameters, even if some of them are optional.

After #697 I'm hoping this will be possible to fix.

pataar:
Closes #736

## Summary

When passing route parameters as an array, omitting optional parameters caused a TypeScript error — even though the parameters were optional. This PR fixes that by making optional parameters truly optional in the generated tuple type.

### Before

Given a route like:

```ts
interface RouteList {
    'posts.comments.show': [
        { name: 'post'; required: true },
        { name: 'comment'; required: false },
    ];
}
```

These perfectly valid calls would all cause type errors:

```ts
route('posts.comments.show', [2]);        // ❌ Source has 1 element(s) but target requires 2
route('posts.comments.show', ['foo']);     // ❌
route('posts.comments.show', [{ id: 2 }]); // ❌
```

The workaround was to always pass all parameters or use object syntax instead.

### After

All of the above now work as expected — no type errors, no workarounds needed:

```ts
route('posts.comments.show', [2]);          // ✅
route('posts.comments.show', ['foo']);       // ✅
route('posts.comments.show', [{ id: 2 }]);  // ✅
route('posts.comments.show', [2, 3]);       // ✅ (still works too)
```

## How it works

The old `KnownRouteParamsArray` type mapped all parameters to required tuple positions regardless of whether they were optional. The fix splits it into two recursive helper types:

- **`RequiredParamsTuple`** — collects required params as required tuple elements
- **`OptionalParamsTuple`** — collects optional params as optional tuple elements (`?`), with `...unknown[]` at the end for extra items

So for the route above, the generated type goes from `[ParameterValue, ParameterValue, ...unknown[]]` to `[ParameterValue, ParameterValue?, ...unknown[]]`.

## Test plan

- Uncommented 4 existing TODO test cases that were previously blocked by this bug
- Added new tests for routes where all parameters are optional (`route('optional', [])` and `route('optional', ['foo'])`)
- All JS and type tests pass

### DIFF
--- a/src/js/index.d.ts
+++ b/src/js/index.d.ts
@@ -114,16 +114,28 @@ type RouteParamsObject<N extends RouteName> = N extends KnownRouteName
 type GenericRouteParamsArray = unknown[];
 /**
  * An array of parameters for a specific named route.
+ *
+ * Required parameters come first as required tuple elements, then optional
+ * parameters become optional tuple elements, followed by arbitrary extras.
  */
 type KnownRouteParamsArray<I extends readonly ParameterInfo[]> = [
-    ...{ [K in keyof I]: Routable<I[K]> },
-    ...unknown[],
+    ...RequiredParamsTuple<I>,
+    ...OptionalParamsTuple<I>,
 ];
-// Because `K in keyof I` for a `readonly` array is always a number, even though this
-// looks like `{ 0: T, 1: U, 2: V }` TypeScript generates `[T, U, V]`. The nested
-// array destructing lets us type the first n items in the array, which are known
-// route parameters, and then allow arbitrary additional items.
-// See https://github.com/tighten/ziggy/pull/664#discussion_r1330002370.
+
+type RequiredParamsTuple<I extends readonly ParameterInfo[]> =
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
Wow........ this is awesome. Thank you so much!

PR #886, opened 2026-05-30
Microbundle to tsdown migration

abetwothree:
This PR implements [tsdown](https://github.com/rolldown/tsdown) to address the vulnerability warnings associated with [microbundler](https://github.com/tighten/ziggy/issues/885).

Functionally, it is the same as microbundler in how it bundles the files.

Long term, it will also allow for the current vanilla JS source files to be TypeScript and have tsdown convert to vanilla JS and generate .d.ts files automatically. That would remove the need for this command to exist `ziggy:generate --types` as types would be picked up automatically.

Thank you for your consideration 😄


bakerkretzmar:
Thanks! Why was `exports.import` added in package.json, what does that do exactly?

I removed a bunch of options that were set to their default values, and added back a shim `index.esm.js` in case people are importing that.

Going to continue testing.

abetwothree:
@bakerkretzmar `exports.import` is for ESM support. `exports.require` is for common js support. 

See the section about "Conditional exports" in the link below
https://nodejs.org/api/packages.html

bakerkretzmar:
Thanks a lot for the PR and appreciate your patience!

PR #888, opened 2026-06-23
Create the directory for a custom types path if it doesn't exist

### LINKED ISSUE #883: ziggy:generate --types=<nested path> fails unless the target directory already exists
I found a CLI issue in `tighten/ziggy` `v2.6.1`.

When using a custom nested path for `--types`, the command writes the file directly but does not ensure the target directory exists.

## Repro

```bash
php artisan ziggy:generate --types=tmp/ziggy-audit/nested/ziggy.d.ts --types-only
```

This fails with:

```text
file_put_contents(.../tmp/ziggy-audit/nested/ziggy.d.ts): Failed to open stream: No such file or directory
```

## Why it happens

`CommandRouteGenerator` ensures the directory for the main JS output path exists, but the custom types path is written later with no corresponding `ensureDirectoryExists()` call.

## Expected

If a custom `--types` path is provided, Ziggy should create its parent directory before writing the file, just like it already does for the main output path.

## Actual

The command fails unless the directory was pre-created manually.

## Relevant code

- `src/CommandRouteGenerator.php`
- main output path setup around the initial `ensureDirectoryExists()` call
- custom `--types` write path later in the command

lazerg:
When a custom path is passed to `--types`, the declaration file is written without first ensuring its parent directory exists. Running `ziggy:generate` with a custom types path therefore fails if that directory isn't already there:

```bash
php artisan ziggy:generate --types-only --types=tmp/ziggy/nested/ziggy.d.ts
# file_put_contents(.../tmp/ziggy/nested/ziggy.d.ts): Failed to open stream: No such file or directory
```

`CommandRouteGenerator` already calls `ensureDirectoryExists()` for the main output path, but not for the custom types path that's written later. This adds the same call before writing the types file, so its directory is created when needed, just like for the main output.

Closes #883.

### DIFF
--- a/src/CommandRouteGenerator.php
+++ b/src/CommandRouteGenerator.php
@@ -55,6 +55,8 @@ class CommandRouteGenerator extends Command
                 default => $this->option('types'),
             };
 
+            $filesystem->ensureDirectoryExists(dirname(base_path($typesPath)), recursive: true);
+
             $filesystem->put(base_path($typesPath), new $types($ziggy));
         }
 
--- a/tests/Unit/CommandRouteGeneratorTest.php
+++ b/tests/Unit/CommandRouteGeneratorTest.php
@@ -163,6 +163,17 @@ test('generate dts file at custom path', function () {
     expect(base_path('resources/js/custom.d.ts'))->toBeFile();
 });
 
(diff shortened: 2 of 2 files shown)


bakerkretzmar:
Thanks!

PR #891, opened 2026-07-01
Honor absolute paths passed to ziggy:generate

### LINKED ISSUE #882: ziggy:generate rebases absolute output paths under base_path() instead of writing to the requested location
I found a path-handling issue in `tighten/ziggy` `v2.6.1`.

`ziggy:generate` appears to wrap both the main output path and the types output path in `base_path(...)` unconditionally.

That means an absolute path like `/tmp/ziggy.js` is treated as relative to the Laravel app root instead of being written to `/tmp/ziggy.js`.

## Repro

```bash
php artisan ziggy:generate /tmp/ziggy-audit-exists/ziggy.js
```

Expected output path:

```text
/tmp/ziggy-audit-exists/ziggy.js
```

Actual output path:

```text
<project root>/tmp/ziggy-audit-exists/ziggy.js
```

## Expected

Absolute paths should be honored as-is.

## Actual

Absolute paths are silently rebased under the application base path.

## Why this matters

This is easy to miss in automation/CI scripts because the command succeeds, but the file is written to the wrong place.

## Relevant code

- `src/CommandRouteGenerator.php`
- the `base_path($path)` / `base_path("{$name}.js")` / `base_path($typesPath)` calls

lazerg:
`ziggy:generate` wraps the output path in `base_path()` unconditionally, so passing an absolute path like `/tmp/ziggy.js` writes the file under the application root instead of where it was requested. This routes every path through a small helper that only prepends `base_path()` when the path is relative, leaving absolute paths untouched.

Closes #882.

### DIFF
--- a/src/CommandRouteGenerator.php
+++ b/src/CommandRouteGenerator.php
@@ -33,10 +33,10 @@ class CommandRouteGenerator extends Command
 
         $path = $this->argument('path') ?? config('ziggy.output.path', 'resources/js/ziggy.js');
 
-        if ($filesystem->isDirectory(base_path($path))) {
+        if ($filesystem->isDirectory($this->path($path))) {
             $path .= '/ziggy';
         } else {
-            $filesystem->ensureDirectoryExists(dirname(base_path($path)), recursive: true);
+            $filesystem->ensureDirectoryExists(dirname($this->path($path)), recursive: true);
         }
 
         $name = preg_replace('/(\.d)?\.ts$|\.js$/', '', $path);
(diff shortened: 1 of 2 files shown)


bakerkretzmar:
I think this is the right move but it's definitely a breaking change so it'll have to go into the next major release. Will try to get it merged sooner than later though and start planning for that. Thanks!

lazerg:
Hey @bakerkretzmar, mind taking a look at this one when you get a chance 🙏

lazerg:
I deleted the fork by accident, which force-closed this PR with no way to reopen it. Restored the branch and reopened it as #895, same commit.

PR #892, opened 2026-07-02
Generate the types file when --types is passed as a boolean

### LINKED ISSUE #879: New --types flag behavior breaks Laravel convention
### Ziggy version

v2.6.1

### Laravel version

v12.53.0

### Description

The new behavior of the --types flag breaks previously working code and the convention described in the [Laravel documentation](https://laravel.com/docs/12.x/artisan#passing-boolean-values) from executing an artisan command programmatically with a boolean flag.

This is the example code:
```php
$exitCode = Artisan::call('migrate:refresh', [
    '--force' => true,
]);
```

When the command is executed like this `true` ends up being coerced into a `1` and that becomes the name of the types file.

### Ziggy call and context

```js
// Called from within another artisan command
$this->call('ziggy:generate', ['--types' => true]);

// This works but is not intuitive
$this->call('ziggy:generate', ['--types' => null]);
```

### Ziggy configuration

```js
n/a
```

### Route definition

```php
n/a
```

lazerg:
Passing `--types` as a boolean (e.g. `Artisan::call('ziggy:generate', ['--types' => true])`, which is the convention Laravel documents for boolean options) coerces the value to the string `"1"`, which then falls through the `match` in `CommandRouteGenerator` and gets used as the types file's path, so the declaration file is written to a file named `1` in the project root instead of at the default location. This adds the coerced boolean value to the arm that already handles `false`/`null`, so a boolean `--types` generates the types file at the default path as expected.

Closes #879.

### DIFF
--- a/src/CommandRouteGenerator.php
+++ b/src/CommandRouteGenerator.php
@@ -51,7 +51,7 @@ class CommandRouteGenerator extends Command
             $types = config('ziggy.output.types', Types::class);
 
             $typesPath = match ($this->option('types')) {
-                'false', null => config('ziggy.output.types-path', "{$name}.d.ts"),
+                'false', '1', true, null => config('ziggy.output.types-path', "{$name}.d.ts"),
                 default => $this->option('types'),
             };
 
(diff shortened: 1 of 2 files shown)


lazerg:
@bakerkretzmar since you already reviewed #891, mind taking a look at this one too?

lazerg:
I deleted the fork by accident, which force-closed this PR with no way to reopen it. Restored the branch and reopened it as #896, same commit.

PR #894, opened 2026-08-17
Check if model has table or route key attribute to determine bindings

bram-pkg:
This boots the model if `#[Table]` or `#[RouteKey]` is used on the model, to allow the model to boot those traits and set the current route key.


bakerkretzmar:
Thanks a lot!

bakerkretzmar:
Doesn't seem to work with traits 🤦 I'll try to get another patch in asap to update that.

bram-pkg:
Thanks! 

What traits do you mean?

bakerkretzmar:
If the attributes are placed on a trait that a model uses, which is valid and works for the routing itself, Ziggy doesn't pick them up. Not your fault just another scenario I want to make sure it covers and didn't want to forget.

bram-pkg:
Ah ok. I hadn't considered that scenario.

